PR #2, opened 2019-02-15
Switch to OpenShift fork of cluster-api.

russellb:
OpenShift has forked the Machine API to have a stable version of the
API to depend on sooner than upstream cluster-api will be ready to
declare one.  We'll target this API first, but aim to somehow support
upstream cluster-api later.


russellb:
This at least compiles.  This still doesn't have any useful functionality in the core of the actuator (pkg/cloud/baremetal/machine/actuator.go).  Next up is to get this running with a custom machine-api-operator.

PR #4, opened 2019-02-21
Add "make build" and dev workflow instructions.

russellb:
This target just simply compiles the two binaries just like the
Dockerfile does.  I left the rest of the existing Makefile, which was
based on the cluster-api provider developer docs for now.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,12 @@
 # Image URL to use all building/pushing image targets
 IMG ?= controller:latest
 
+.PHONY: build
+build:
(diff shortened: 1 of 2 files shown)


dhellmann:
LGTM, but I haven't tested it. Do you want anyone specific to look at this before we merge it?

russellb:
Not really, we can merge and fix with follow-up PRs.  It's got to be better than nothing.  :-)

russellb:
I'll merge based on your LGTM

PR #6, opened 2019-02-21
Sync "Add directives for setting ClusterRole permissions"

russellb:
This actuator was based on docs in the upstream cluster-api gitbook.
This change syncs a change from those docs.

https://github.com/kubernetes-sigs/cluster-api/pull/750/files


dhellmann:
LGTM. Does kubebuilder interpret those comments in some way?

russellb:
I think so!  I honestly haven't figured out how all of this works yet, just trying to stay in sync with the documented starting point while I continue figuring it out ...

PR #7, opened 2019-02-21
Fix compatibility with newer kustomize.

russellb:
The newest releases of kustomize block relative paths in the config.  This patch pulls in a fix to our vendored copy of cluster-api, and fixes our own config as well.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,7 @@ deploy: manifests
 # Generate manifests e.g. CRD, RBAC etc.
 manifests:
 	go run vendor/sigs.k8s.io/controller-tools/cmd/controller-gen/main.go all
-	kustomize build config/default/ > provider-components.yaml
+	kustomize build config/ > provider-components.yaml
 	echo "---" >> provider-components.yaml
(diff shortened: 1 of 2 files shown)


russellb:
Ok, thanks for not letting me cheat.  :-)

I was hoping it was super short term and I was being lazy about it.  It sounds like the right thing is for me to create my own fork to use as long as the patch is needed.  Thanks.

russellb:
... ok I've decided that switching to a temporary fork is more of a pain than it's worth.  This issue can worked around in the meantime by using an older version of kustomize.

mhrivnak:
> ... ok I've decided that switching to a temporary fork is more of a pain than it's worth. This issue can worked around in the meantime by using an older version of kustomize.

That's what I would recommend. It's working well for me.

russellb:
I've updated this PR since the necessary change merged into openshift/cluster-api

russellb:
I resolved the conflicts here, but kustomize-2 won't actually work yet until we update our vendored copy of `sigs.k8s.io/cluster-api`

russellb:
#37 updated to latest cluster-api, so this can merge now

PR #8, opened 2019-02-22
adds basic actuator implementation of create and update

mhrivnak:
The actuator "creates" a machine by adding a MachineRef value to a
BareMetalHost. That value tells the BareMetalOperator to provision the host and
turn it into a Node by way of the Machine.


Gopkg.toml
```diff
@@ -20,6 +20,10 @@ required = [
   go-tests = true
 
 
+[[constraint]]
+  name="github.com/metalkube/baremetal-operator"
+  branch="master"
+
```
russellb: Can we put the addition of the new vendored code in its own commit?

mhrivnak: Maybe, but `dep` will complain that the code is listed as a constraint but not used. There's probably a way to force it, but that would be an unusual approach. The prevailing technique for vendoring with `dep` is to add dependent code with the code that uses it.

What's the motivation for separate commits?

pkg/cloud/baremetal/actuators/machine/actuator.go
```diff
@@ -66,13 +109,46 @@ func (a *Actuator) Delete(ctx context.Context, cluster *machinev1.Cluster, machi
...
-	return fmt.Errorf("TODO: Not yet implemented")
+	host, err := a.getHost(ctx, machine)
+	if err != nil {
+		return err
+	}
+	if host == nil {
+		return fmt.Errorf("host not found for machine %s", machine.Name)
+	}
+
+	changed, err := ensureLabelOnHost(ctx, machine, host)
```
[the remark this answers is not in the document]
mhrivnak: They're changing different objects. On a Machine, the only thing we ever change right now is adding the annotation. So it seems reasonable for that function to have the update call in one place.

On the BareMetalHost, there may or may not be other changes to the host that should be committed at the same time as adding the label, so in that case it seems reasonable to have it not do any updating on its own. The downside is that the update call appears in multiple places, wherever this function gets called.

I'm not tied to any of that and am happy to make sure it's clear and consistent. Let me know what you think.

pkg/cloud/baremetal/actuators/machine/actuator.go
```diff
@@ -90,3 +166,96 @@ func (a *Actuator) GetKubeConfig(cluster *machinev1.Cluster, controlPlaneMachine
...
+
+// chooseHost iterates through known hosts and returns one that can be
+// associated with the machine. It searches all hosts in case one already has an
+// association with this machine, but lacks the label that would have made it
+// discoverable. It will add a Machine reference before returning the host.
+func (a *Actuator) chooseHost(ctx context.Context, machine *machinev1.Machine) (*bmh.BareMetalHost, error) {
+	// get list of BMH
+	hosts := bmh.BareMetalHostList{}
+	opts := &client.ListOptions{}
+	a.client.List(ctx, opts, &hosts)
```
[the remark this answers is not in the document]
mhrivnak: Agreed. We need to work out exactly how the BMH state will be represented through its whole lifecycle and then plug that knowledge into this controller.

pkg/cloud/baremetal/actuators/machine/actuator.go
```diff
@@ -90,3 +166,96 @@ func (a *Actuator) GetKubeConfig(cluster *machinev1.Cluster, controlPlaneMachine
...
+			chosenHost = &host
+			break
+		}
+	}
+	if chosenHost == nil {
+		if len(availableHosts) == 0 {
+			return nil, fmt.Errorf("No available host found")
+		}
+		// choose a host at random from available hosts
+		rand.Seed(time.Now().Unix())
```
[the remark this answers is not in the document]
mhrivnak: Probably. I just went with random selection so that if you have a pool of machines being added and removed (not sure how much churn is realistic to expect), they're not always selected in the same order, leaving the first one busy all the time, and the last one perhaps idle forever as a cold spare. But that does make testing perhaps a bit harder, although that feels like the kind of lack-of-determinism that can be inherent to declarative APIs anyway.

pkg/cloud/baremetal/actuators/machine/actuator.go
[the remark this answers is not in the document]
dhellmann: Oh, duh. I misread the docstring on the const definitions and obviously didn't look closely enough here to notice the different objects in use.

Gopkg.toml
russellb: Cleaner git history.  The commit with the real code in it will be much easier to read without the vendored code included.  I know github helps during review, but not when looking back later via other tools, such as the git command line.

I would imagine having all of the changes staged in your local tree such that dep doesn't complain.  To commit them, "git add vendor ; git commit -m 'vendor: add vendored whatever' ; git commit -a 'the real code changes'"

russellb: I like having it in the same PR, just not the same commit.

I saw this advice in this doc recently, too: https://github.com/openshift/installer/blob/master/docs/dev/dependencies.md

That doc looks like a useful thing we could link to from our own contributing doc, when we start one.  Maybe a general metalkube contributors guide would be good for metalkube-docs?  This is something we could link to as misc advice.

mhrivnak: Ah, I see. Let me give it a shot...

mhrivnak: done

russellb: Thanks!

pkg/cloud/baremetal/actuators/machine/actuator.go
```diff
@@ -90,3 +166,96 @@ func (a *Actuator) GetKubeConfig(cluster *machinev1.Cluster, controlPlaneMachine
...
+	}
+	switch len(hosts.Items) {
+	case 0:
+		return nil, nil
+	case 1:
+		host := &hosts.Items[0]
+		log.Printf("Found associated host %s", host.Name)
+		return host, nil
+	default:
+		return nil, fmt.Errorf("Found %d hosts with label %s:%s", len(hosts.Items), "machine", machine.Name)
```
russellb: Should this be a TODO item?  I feel like we should be trying to recover from this, by choosing which Host we really want to keep, and disassociating from any others?  That could be in a later PR though.

pkg/cloud/baremetal/actuators/machine/actuator.go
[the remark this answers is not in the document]
russellb: Maybe add a TODO comment in the code here?

// TODO - This is where we should refine our query criteria to filter out hosts that are invalid due to some criteria, like operational status, ...

russellb:
LGTM.  This looks like an awesome start to the actuator!  I'm find merging this whenever @dhellmann is good with it, too.

pkg/cloud/baremetal/actuators/machine/actuator.go
dhellmann: I like the idea of leaving TODO and FIXME comments as we continue to build this out. Let's do that in a future update?

russellb: TODO comments proposed in PR #9

PR #16, opened 2019-03-04
Removes label creation on BareMetalHost and adds actuator unit tests

### LINKED ISSUE #11: Use annotation to identify the associated BareMetalHost
The actuator currently uses a label on the BareMetalHost as the primary way of identifying which host is associated with a Machine.

This should be changed such that the primary source is an annotation on the Machine.  If the annotation is not present for some reason, it should iterate over BareMetalHosts looking for a host with a matching MachineRef.

mhrivnak:
fixes #11

### DIFF
--- a/pkg/cloud/baremetal/actuators/machine/actuator.go
+++ b/pkg/cloud/baremetal/actuators/machine/actuator.go
@@ -26,7 +26,7 @@ import (
 	bmh "github.com/metalkube/baremetal-operator/pkg/apis/metalkube/v1alpha1"
 	machinev1 "github.com/openshift/cluster-api/pkg/apis/machine/v1beta1"
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/api/errors"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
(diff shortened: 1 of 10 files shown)


pkg/cloud/baremetal/actuators/machine/actuator.go
```diff
@@ -167,36 +144,37 @@ func (a *Actuator) GetKubeConfig(cluster *machinev1.Cluster, controlPlaneMachine
...
+	if annotations == nil {
+		return nil, nil
 	}
-	err := a.client.List(ctx, opts, &hosts)
-	if err != nil {
-		return nil, err
+	hostName, ok := annotations[HostAnnotation]
+	if !ok {
+		return nil, nil
+	}
```
russellb: In our reconcile design notes, we had written that in the case of an annotation not being present, we'd also look through the BareMetalHosts looking for a machine with a matching machineRef.

That would be an unexpected case, but I suppose the actuator needs to recover from it somehow.

Do you want to add code for that in this PR?  add a TODO in the code?  I noticed we didn't create an issue for that bit.

russellb: ... nevermind, as soon as I hit enter, I see that this case is already covered in chooseHost().  :-)

PR #21, opened 2019-03-07
Limits chooseHost to hosts in the machine's own namespace

mhrivnak:
Also includes minor refactor improvements

### DIFF
--- a/pkg/cloud/baremetal/actuators/machine/actuator.go
+++ b/pkg/cloud/baremetal/actuators/machine/actuator.go
@@ -195,10 +195,11 @@ func (a *Actuator) chooseHost(ctx context.Context, machine *machinev1.Machine) (
 	hosts := bmh.BareMetalHostList{}
 	// TODO - We should add filtering here for known conditions that make
 	// a host not a valid choice, such as an operational state of error.
-	opts := &client.ListOptions{}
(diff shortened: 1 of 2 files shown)


dhellmann:
Why do we want to do this? Are we sure we're going to run all of these components in the same namespace?

mhrivnak:
> Why do we want to do this? Are we sure we're going to run all of these components in the same namespace?

Just one quick clarification: where these controllers run is not necessarily the same place the Machine and BareMetalHost objects live. This PR is assuming the hosts and machines themselves will be in the same namespace, but makes no assumptions about where the controllers run.

Searching all namespaces here is not a good plan, so we definitely want to limit this query to one namespace. So far I haven't seen any reasons to put machines and hosts in separate namespaces, but putting them in the same is convenient for RBAC. So I think that's a reasonable starting point. If we find a reason to put them in separate namespaces, then we'll need to give the actuator some way to figure out what the host namespace is and pass it into this code. It would be a fairly simple addition. So I think this is the right approach for now, but we can add the ability to do separate namespaces later if needed.

dhellmann:
OK. This feels like a good policy/assumption/rule/whatever to document with a design doc in the metalkube-docs repo, especially if it's going to have implications on how users interact with the system that we will need to put into user documentation later.

PR #23, opened 2019-03-11
uses BareMetalHost.Available method to determine if a host is available

### LINKED ISSUE #12: Match against Hosts ready to be provisioned
Related to https://github.com/metalkube/baremetal-operator/issues/31

Once BareMetalHost provides an easy way to determine whether a given host is ready to be provisioned, it should be used as part of the Host selection criteria in the Machine actuator.

mhrivnak:
fixes #12


russellb:
lgtm

PR #24, opened 2019-03-13
Add node addresses to Machine status

bcrochet:
Eventually, this will add the node address(es) to the status of the
Machine object. For now, it's just an empty list.

### DIFF
--- a/pkg/cloud/baremetal/actuators/machine/actuator.go
+++ b/pkg/cloud/baremetal/actuators/machine/actuator.go
@@ -26,7 +26,9 @@ import (
 	bmh "github.com/metal3-io/baremetal-operator/pkg/apis/metal3/v1alpha1"
 	bmv1alpha1 "github.com/metal3-io/cluster-api-provider-baremetal/pkg/apis/baremetal/v1alpha1"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/tools/cache"
 	"sigs.k8s.io/cluster-api/pkg/apis/cluster/common"
 	machinev1 "sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha1"
@@ -123,6 +125,10 @@ func (a *Actuator) Create(ctx context.Context, cluster *machinev1.Cluster, machi
 		return err
 	}
 
+	if err := a.updateMachineStatus(ctx, machine, host); err != nil {
+		return err
+	}
+
 	log.Printf("Finished creating machine %v .", machine.Name)
 	return nil
 }
@@ -174,6 +180,11 @@ func (a *Actuator) Update(ctx context.Context, cluster *machinev1.Cluster, machi
 	if err != nil {
 		return err
 	}
+
+	if err := a.updateMachineStatus(ctx, machine, host); err != nil {
+		return err
+	}
+
 	log.Printf("Finished updating machine %v .", machine.Name)
 	return nil
 }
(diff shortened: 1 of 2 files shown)


mhrivnak:
Just for context, what specifically need to read these addresses off of the Machine?

mhrivnak:
Do you think this controller will need to Watch BareMetalHost events? I assume that's the only way that this controller will be able to update the Machine status in response to host changes.

pkg/cloud/baremetal/actuators/machine/actuator.go
```diff
@@ -25,8 +25,11 @@ import (
 
 	bmh "github.com/metalkube/baremetal-operator/pkg/apis/metalkube/v1alpha1"
 	machinev1 "github.com/openshift/cluster-api/pkg/apis/machine/v1beta1"
+	clusterclient "github.com/openshift/cluster-api/pkg/client/clientset_generated/clientset"
```
russellb: This will need to be rebased on top of #26

mhrivnak: Sorry to pile on your WIP, but I'll just quickly point out one thing before this goes farther: we should use the controller-runtime client (the actuator's `client` attribute) instead of this generated clientset so that we can take advantage of automatic caching.

bcrochet: > Just for context, what specifically need to read these addresses off of the Machine?

The coredns plugin will be using the info in the Machine object.

bcrochet:
> Is there benefit to merging this before we're able to populate it with real addresses?

I don't think so. It would be a huge noop atm. But I'm not opposed.

mhrivnak:
It seems that these addresses get populated onto the Node resource already. Could you retrieve them from there? Or do they need to be on the Machine resource for some reason?

bcrochet:
> It seems that these addresses get populated onto the Node resource already. Could you retrieve them from there? Or do they need to be on the Machine resource for some reason?

They should be on the Machine resource as well. Other CAP operate this way. For one thing, it would be for consistency.

mhrivnak:
Following up to add context: As far as I can tell, there are only two other actuators that set `Addresses` on a machine's status.

* One is `openshift/cluster-api-provider-aws`. Notably, its upstream `kubernetes-sigs/cluster-api-provider-aws` does not set `Addresses`, so that seems to be openshift-specific.
* The other is `openshift/cluster-api-provider-libvirt`, which does not have an upstream in kubernetes-sigs.

None of the non-openshift actuators seem to set this field.

It would be helpful to understand why it is useful to read these IPs from a Machine instead of from a Node.

russellb:
> Following up to add context: As far as I can tell, there are only two other actuators that set `Addresses` on a machine's status.
> 
> * One is `openshift/cluster-api-provider-aws`. Notably, its upstream `kubernetes-sigs/cluster-api-provider-aws` does not set `Addresses`, so that seems to be openshift-specific.
> * The other is `openshift/cluster-api-provider-libvirt`, which does not have an upstream in kubernetes-sigs.
> 
> None of the non-openshift actuators seem to set this field.
> 
> It would be helpful to understand why it is useful to read these IPs from a Machine instead of from a Node.

AFAICT the value is in the CSR machine-approver.  It approves a cert request if it finds a Machine with matching addresses to the ones on a Node.  It's using this as a sort of sanity check that the Node is one that's supposed to be allowed to join the cluster.

bcrochet:
Another step towards openshift-metal3/dev-scripts#260

Gopkg.lock
```diff
@@ -314,6 +314,14 @@
   revision = "65fb64232476ad9046e57c26cd0bff3d3a8dc6cd"
   version = "v1.4.3"
 
+[[projects]]
+  digest = "1:ef04c20bf6f1dcd5e1cd1fd27a65675649dfed58e89699fe2891cccc0a233efd"
+  name = "github.com/openshift/cluster-api"
```
russellb: We shouldn't be pulling in the openshift cluster-api fork.  This repo should only use the upstream cluster-api, which is already vendored

Gopkg.lock
```diff
@@ -462,12 +470,12 @@
   version = "v1.9.1"
 
 [[projects]]
-  branch = "release-branch.go1.11"
-  digest = "1:6c1f398ca39ca7b372787df7833b71af914021af747e83c17163bad6aee25594"
+  branch = "master"
+  digest = "1:287367cdf1d4ad1b717dcc43a31f500eb36a2bdfc1ad3305fc023e58e1ece269"
```
russellb: Was this necessary, or pulled in by a broad `'dep ensure`?  Can you limit the PR to the changes necessary for the feature being added?

pkg/cloud/baremetal/actuators/machine/actuator.go
russellb: @bcrochet looks like this comment still needs to be addressed

russellb: (the one from Michael)

russellb:
> Do you think this controller will need to Watch BareMetalHost events? I assume that's the only way that this controller will be able to update the Machine status in response to host changes.

That would indeed be an enhancement.  Right now though, all of this data will be present before a Host is chosen for a Machine, and then will never change.  It's possible that we will start updating the data on the host in the future.  At that point we'll need to watch the Host for changes.

in short, i think we're ok not watching it right now, but we probably will need to in the future

pkg/cloud/baremetal/actuators/machine/actuator.go
```diff
@@ -370,3 +381,54 @@ func configFromProviderSpec(providerSpec machinev1.ProviderSpec) (*bmv1alpha1.Ba
...
+	now := metav1.Now()
+	machineCopy.Status.LastUpdated = &now
+
+	err := a.client.Status().Update(ctx, machineCopy)
+	return err
+}
+
+// NodeAddresses returns a slice of corev1.NodeAddress objects for a
+// given Baremetal machine.
+func (a *Actuator) NodeAddresses(host *bmh.BareMetalHost) ([]corev1.NodeAddress, error) {
```
russellb: Does something outside of this package use NodeAddresses?  or should it be nodeAddresses?

bcrochet: Nope. Switched to nodeAddresses.

bcrochet:
> Can you add some test coverage for this in actuator_test.go ?

I'll work on that.

russellb:
Thanks for adding the tests!

I just gave this a try in `metal3-dev-env` and it worked as expected.  Thanks!

PR #26, opened 2019-03-14
Switch to sigs.k8s.io/cluster-api

### LINKED ISSUE #25: Switch to upstream cluster-api
This repository currently uses openshift/cluster-api.  We need to fix that and move to the upstream kubernetes-sigs/cluster-api repository instead.

The openshift/cluster-api-provider-baremetal fork can contain the changes necessary to use OpenShift's copy.

russellb:
This repo was previously using github.com/openshift/cluster-api.  We
should be using the upstream version of cluster-api, and not the
OpenShift fork.

OpenShift has a fork of this repository where it can maintain changes
necessary to switch back to their copy of cluster-api.

Fixes #25.


russellb:
@mhrivnak dep made a lot more changes to Gopkg.lock than I expected.   Can you look over that for me?  Maybe it's the version of dep I'm using compared to what generated it last?

mhrivnak:
> @mhrivnak dep made a lot more changes to Gopkg.lock than I expected. Can you look over that for me? Maybe it's the version of dep I'm using compared to what generated it last?

Did you run `dep ensure -update` by chance without specifying a particular dependency?

russellb:
Yes, I did.  I wanted to make sure it pruned the stuff that was no longer needed, so I just ran "dep ensure".

mhrivnak:
Ok, I'm guessing you have an older version of dep, perhaps prior to 0.5.0. When I check out your branch, grab the old Gopkg.lock, then run `dep ensure`. it keeps all the "pruneopts" and digest values that your branch removes. It's probably a good idea to update your dep install and run it again.

russellb:
Ah ha, good catch.  I thought I had a new dep, but I also had an old system packaged version, and that's what got used.

russellb:
@mhrivnak fixed.  looks better now!

PR #28, opened 2019-03-14
uses k8s.io/client-go/tools/cache to generate host annotation value

mhrivnak:
I noticed that ReconcileMachineHealthCheck in the machine-api-operator uses
this approach when it puts a machine annotation on a Node. It seems worth using
the same approach for the host annotation.

### DIFF
--- a/pkg/cloud/baremetal/actuators/machine/actuator_test.go
+++ b/pkg/cloud/baremetal/actuators/machine/actuator_test.go
@@ -152,7 +152,7 @@ func TestExists(t *testing.T) {
 			Machine: machinev1.Machine{
 				ObjectMeta: metav1.ObjectMeta{
 					Annotations: map[string]string{
-						HostAnnotation: "somehost",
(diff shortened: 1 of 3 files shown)


pkg/cloud/baremetal/actuators/machine/actuator_test.go
```diff
@@ -393,7 +398,7 @@ func TestDelete(t *testing.T) {
 					Name:      "mymachine",
 					Namespace: "myns",
 					Annotations: map[string]string{
-						HostAnnotation: "myhost",
+						HostAnnotation: "myhost/myns",
```
russellb: This and the rest in this file look backwards to me.  Should it be "myns/myhost" and this test just isn't checking it?

mhrivnak: doh, good catch. Apparently I need to improve the test to catch that!

PR #38, opened 2019-03-20
WIP - Add clusterctl.

russellb:
I noticed that other providers include a build of clusterctl.  We
don't support create Clusters yet, but we can at least make sure this
builds.


russellb:
Well, this built when I first posted this, but doesn't build anymore.  Marking WIP for now.

cmd/clusterctl/main.go
```diff
@@ -0,0 +1,26 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	_ "sigs.k8s.io/cluster-api-provider-openstack/pkg/cloud/openstack"
```
russellb: copy / paste error here

russellb:
i'll just reopen this later when it's relevant

PR #39, opened 2019-03-22
Klog flags

vrutkovs:
This should resolve `flag provided but not defined: -logtostderr` error when clusterapi start machine controller

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -24,6 +24,7 @@ import (
 	bmoapis "github.com/metalkube/baremetal-operator/pkg/apis"
 	"github.com/metalkube/cluster-api-provider-baremetal/pkg/apis"
 	"github.com/metalkube/cluster-api-provider-baremetal/pkg/cloud/baremetal/actuators/machine"
+	"k8s.io/klog"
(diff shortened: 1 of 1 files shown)


russellb:
This code was based on the docs from https://github.com/kubernetes-sigs/cluster-api/tree/master/docs/book/provider_implementations

You may want to consider proposing this addition there, too.

vrutkovs:
Filed https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/271 to track this. I'm not entirely sure if this PR would fix it, so didn't propose upstream PR yet

russellb:
I tried this fix, but it doesn't seem to work for me.

```
bin/machine-controller-manager --logtostderr=true --v=3
flag provided but not defined: -logtostderr
```

cmd/manager/main.go
```diff
@@ -33,18 +34,30 @@ import (
 )
 
 func main() {
```
russellb: This patch works if I add this to the beginning of main():

```
    klog.InitFlags(nil)
```

vrutkovs: Awesome, thanks!

PR #43, opened 2019-03-25
enables unit test in CI

mhrivnak:
The bug that caused us to turn this off has been resolved.

### DIFF
--- a/tools/install_kubebuilder.sh
+++ b/tools/install_kubebuilder.sh
@@ -1,0 +1,11 @@
+#!/usr/bin/env bash
+
+version=1.0.8
+arch=amd64
+
+curl -L -O "https://github.com/kubernetes-sigs/kubebuilder/releases/download/v${version}/kubebuilder_${version}_linux_${arch}.tar.gz"
+
+tar -zxvf kubebuilder_${version}_linux_${arch}.tar.gz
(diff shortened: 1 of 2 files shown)


tools/install_kubebuilder.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+version=1.0.8
+arch=amd64
+
+curl -L -O "https://github.com/kubernetes-sigs/kubebuilder/releases/download/v${version}/kubebuilder_${version}_linux_${arch}.tar.gz"
+
+tar -zxvf kubebuilder_${version}_linux_${arch}.tar.gz
+mv kubebuilder_${version}_linux_${arch} kubebuilder && sudo mv kubebuilder /usr/local/
```
dhellmann: Does something expect it to be in /usr/local, or is it sufficient to add the unpacked directory to $PATH?

mhrivnak: It seems to ignore `$PATH` and is hard-coded somewhere to look in `/usr/local/kubebuilder/bin/`. I'll dig and see if there are any options.

dhellmann: Meh, this is fine. I just like to double check when I see sudo in a CI script.

PR #44, opened 2019-03-25
installs kubebuilder locally instead of globally

mhrivnak:


dhellmann:
This does look better.

PR #50, opened 2019-04-03
do not trigger provisioning when selecting hosts for masters

dhellmann:
Assume a master has already been provisioned and do not set the image
or user data so we avoid triggering provisioning.

### DIFF
--- a/pkg/cloud/baremetal/actuators/machine/actuator.go
+++ b/pkg/cloud/baremetal/actuators/machine/actuator.go
@@ -43,6 +43,8 @@ const (
 	instanceImageSource      = "http://172.22.0.1/images/redhat-coreos-maipo-latest.qcow2"
 	instanceImageChecksumURL = instanceImageSource + ".md5sum"
 	requeueAfter             = time.Second * 30
+
+	machineRoleLabel = "sigs.k8s.io/cluster-api-machine-role"
 )
 
 // Add RBAC rules to access cluster-api resources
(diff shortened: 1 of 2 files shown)


dhellmann:
This is based on our discussion from earlier in the week, but now that I have it worked out I don't know how important it is. If we assume we're going to import the master hosts and machines pre-populated, we shouldn't need to worry about provisioning being triggered.

mhrivnak:
I had this thought while starting to review this PR:

We've talked about specifying the image in the provider spec on the machine. That way it can be defined on the MachineSet, and that gives a reasonable API for the installer or some other actor to set and/or change the value.

Maybe that's where this decision should happen. Rather than hard-coding that machines with the particular label should not have their image set, what do you think about letting that decision be made by whatever creates the Machines for masters? They would just leave that field of the provider spec empty. Then if someone has other machines they don't want provisioned, this would support that. Or if someone figures out how to provision masters using the machine-api, this would already work for them. Thoughts?

dhellmann:
> I had this thought while starting to review this PR:
> 
> We've talked about specifying the image in the provider spec on the machine. That way it can be defined on the MachineSet, and that gives a reasonable API for the installer or some other actor to set and/or change the value.
> 
> Maybe that's where this decision should happen. Rather than hard-coding that machines with the particular label should not have their image set, what do you think about letting that decision be made by whatever creates the Machines for masters? They would just leave that field of the provider spec empty. Then if someone has other machines they don't want provisioned, this would support that. Or if someone figures out how to provision masters using the machine-api, this would already work for them. Thoughts?

I like that. Where does that happen? Do we have any of that implemented, yet?

mhrivnak:
> I like that. Where does that happen? Do we have any of that implemented, yet?

Not yet. We need to implement our own provider spec. I've seen some actuators have a "provider config" instead but I haven't looked closely to see what the difference is. Either way, something in the installer is going to create a MachineSet, and that's the thing that will need to have information like this.

dhellmann:
I'll abandon this one and we can write another patch when we're ready to do this with values coming from the machine set.

PR #59, opened 2019-04-18
Adds ProviderSpec with settings for image and userdata.

### LINKED ISSUE #31: actuator sets config drive secret on BareMetalHost
A "config drive" Secret will be present in the cluster, and it will contain data that the BareMetalHost needs in order to configure a provisioned host. For now the secret reference is always the same, so it can be hard-coded. In the future the data may be read from the Machine's provider spec.

This task is to ensure the actuator writes the values to corresponding fields on the BareMetalHost.

As an example, currently in openshift the secret that contains worker config data is showing up with this identity:

```
  name: worker-user-data
  namespace: openshift-machine-api
```

mhrivnak:
The values passed in with the ProviderSpec were previously hard-coded. Any user
of Machines or MachineSets with this controller will need to start including a
suitable ProviderSpec.

fixes #31


mhrivnak:
This cannot be merged until we fix the installer to include a suitable `ProviderSpec`. I'd also like to add a few more tests before merging, but feedback would be helpful in the meantime.

config/crds/baremetal_v1alpha1_baremetalmachineproviderspec.yaml
```diff
@@ -19,17 +19,28 @@ spec:
             of an object. Servers should convert recognized schemas to the latest
             internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#resources'
           type: string
+        imageChecksumURL:
```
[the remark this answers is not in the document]
mhrivnak: Good idea. I'll see how it looks.

It's tempting to use the BMO's Image struct directly, but since these are strictly-versioned APIs (or will be when they graduate from alpha), I think I'll keep the implementations separate and just copy the values into the BMH Spec. That way a change in the BMO's CRD won't necessarily cause a change to the ProviderSpec here.

README.md
```diff
@@ -10,3 +10,28 @@ For more information about this actuator and related repositories, see
...
+metadata:
+  labels:
+    controller-tools.k8s.io: "1.0"
+  name: sample0
+spec:
+  providerSpec:
+    value:
+      apiVersion: "baremetal.cluster.k8s.io/v1alpha1"
+      kind: "BareMetalMachineProviderSpec"
+      imageURL: "http://172.22.0.1/images/rhcos-ootpa-latest.qcow2"
```
dhellmann: The example here doesn't match the code now.

mhrivnak: Fixed.

russellb:
> This cannot be merged until we fix the installer to include a suitable `ProviderSpec`. I'd also like to add a few more tests before merging, but feedback would be helpful in the meantime.

On second thought, I don't think we have to worry about this.  It doesn't affect OpenShift until `openshift/cluster-api-provider-baremetal` merges these changes.

We can go ahead and merge here

config/crds/baremetal_v1alpha1_baremetalmachineproviderspec.yaml
```diff
@@ -19,17 +19,34 @@ spec:
...
-          type: object
-        status:
+        userData:
+          description: UserData references the Secret that holds user data needed
+            by the bare metal operator. The Namespace is optional; it will default
+            to the Machine's namespace if not specified.
           type: object
+      required:
+      - image
+      - userData
```
russellb: I actually think userData should be optional.  We can't provision a host at all without an image, but it's possible that the image we're provisioning has some other way to bootstrap itself that doesn't need user data.

russellb:
Here's a branch with a rebase plus a suggested change to make userdata optional.  Feel free to squash it in if you agree with it.

https://github.com/russellb/cluster-api-provider-baremetal/tree/pr-59

PR #69, opened 2019-05-07
Use metal3 namepsace for cluster-api components.

russellb:
Use the metal3 namespace for the components built from the cluster-api
repository, as well.


russellb:
Michael was sitting next to me while I did this and verbally approved of the approach, so I'm merging :)

PR #73, opened 2019-05-13
Reflect that userdata is optional in tests.

russellb:
We made a late change to the providerSpec to specify that the userData
field was optional.  This patch updates the test code to reflect this,
as well.

I also spotted one other place that assumed userData was specified in
the machine actuator.


russellb:
This patch went in as part of another PR -- 37770fcbffad9543a659aa18ab5e9b5ede03969c

PR #74, opened 2019-05-14
Add BareMetalHost selection criteria using hostSelector.

### LINKED ISSUE #66: Add support for a hostSelector field for specifying BareMetalHost selection criteria
Particularly with bare metal hosts, there will be cases where administrators would like to provision a specific `BareMetalHost` instance.  Our current behavior is simple: choose any available `BareMetalHost`.

To allow some selection criteria, we should add a `hostSelector` field (or something similar) which allows a selector expression on labels of BareMetalHost objects.

https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/

This selector can either be specified directly on a Machine, or a MachineSet. This needs to reside in our providerSpec, so this PR is a prerequisite: https://github.com/metal3-io/cluster-api-provider-baremetal/pull/59

russellb:
This patch series introduces a new field to the providerSpec: `hostSelector`.

The `hostSelector` field can be used to specify label matching criteria against `BareMetalHost` objects, giving control over which hosts are considered valid choices for a given `Machine`.

Usage is outlined in docs/api.md.

Closes #66.


docs/api.md
```diff
@@ -70,3 +81,84 @@ metadata:
...
+* **!** -- Key does not exist.  Values ignored.
+* **=** -- Key equals specified value.  There must only be one
+  value specified.
+* **==**  -- Key equals specified value.  There must only be one
+  value specified.
+* **in**  -- Value is a member of a set of possible values
+* **!=**  -- Key does not equal the specified value.  There must
+  only be one value specified.
+* **notin**  -- Value not a member of the specified set of values.
+* **exists**  - Key exists.  Values ignored.
```
rustyrobot: nit: s/-/--/

russellb: Thanks!  Fixed

pkg/apis/baremetal/v1alpha1/baremetalmachineproviderspec_types.go
```diff
@@ -40,6 +41,28 @@ type BareMetalMachineProviderSpec struct {
...
+	// HostSelector specifies matching criteria for labels on BareMetalHosts.
+	// This is used to limit the set of BareMetalHost objects considered for
+	// claiming for a Machine.
+	HostSelector HostSelector `json:"hostSelector,omitempty"`
+}
+
+// HostSelector specifies matching criteria for labels on BareMetalHosts.
+// This is used to limit the set of BareMetalHost objects considered for
+// claiming for a Machine.
+type HostSelector struct {
```
dhellmann: I'm surprised these data types have to be defined here. Are there not shareable versions of the structs defined somewhere?

pkg/cloud/baremetal/actuators/machine/actuator.go
```diff
@@ -245,34 +248,65 @@ func (a *Actuator) getHost(ctx context.Context, machine *machinev1.Machine) (*bm
...
 		Namespace: machine.Namespace,
 	}
+
 	err := a.client.List(ctx, opts, &hosts)
 	if err != nil {
 		return nil, err
 	}
 
+	// Using the label selector on ListOptions above doesn't seem to work.
+	// I think it's because we have a local cache of all BareMetalHosts.
```
dhellmann: I've seen different behavior in the unit test client and the real thing. Did it not work in tests, or on a real cluster?

pkg/cloud/baremetal/actuators/machine/actuator.go
```diff
@@ -245,34 +248,65 @@ func (a *Actuator) getHost(ctx context.Context, machine *machinev1.Machine) (*bm
...
+		reqs = append(reqs, *r)
+	}
+	labelSelector = labelSelector.Add(reqs...)
+
 	availableHosts := []*bmh.BareMetalHost{}
 
 	for i, host := range hosts.Items {
 		if host.Available() {
-			availableHosts = append(availableHosts, &hosts.Items[i])
+			if labelSelector.Matches(labels.Set(host.ObjectMeta.Labels)) {
```
dhellmann: How does this work if there are no rules specified? Does labelSelector.Matches() default to true if it has no rules?

pkg/cloud/baremetal/actuators/machine/actuator.go
russellb: I tried it on both and got the same behavior.  I didn't spend too much time digging into it, though TBH.

pkg/cloud/baremetal/actuators/machine/actuator.go
russellb: That's right - it defaults to true if there are no requirements specified.

Here's the relevant code: https://github.com/metal3-io/cluster-api-provider-baremetal/blob/master/vendor/k8s.io/apimachinery/pkg/labels/selector.go#L337-L347

pkg/apis/baremetal/v1alpha1/baremetalmachineproviderspec_types.go
russellb: I don't think so.  There's a Requirement type that looks just like HostSelectorRequirement in here:

https://github.com/metal3-io/cluster-api-provider-baremetal/blob/master/vendor/k8s.io/apimachinery/pkg/labels/selector.go#L109-L121

but I think I have to define it here to add the right type annotations so that the CRD lists all of the fields.  If I drop the `json` annotations for HostSelectorRequirement members, for example, then MatchExpressions i just listed as an opaque object in the CRD definition.

russellb:
> This looks good, but needs a rebase.

Thanks!  I've rebased it and will merge.

Issue #14, opened 2019-03-04
tests fail


mhrivnak:
I'm seeing this on the master branch.

```
$ make test
go generate ./pkg/... ./cmd/...
go fmt ./pkg/... ./cmd/...
go vet ./pkg/... ./cmd/...
go run vendor/sigs.k8s.io/controller-tools/cmd/controller-gen/main.go crd
CRD files generated, files can be found under path /home/mhrivnak/golang/src/github.com/metalkube/cluster-api-provider-baremetal/config/crds.
kustomize build config/default/ > provider-components.yaml
2019/03/04 10:16:12 Adding nameprefix and namesuffix to Namespace resource will be deprecated in next release.
echo "---" >> provider-components.yaml
cd vendor && kustomize build github.com/openshift/cluster-api/config/default/ >> ../provider-components.yaml
2019/03/04 10:16:21 Adding nameprefix and namesuffix to Namespace resource will be deprecated in next release.
go test ./pkg/... ./cmd/... -coverprofile cover.out
?   	github.com/metalkube/cluster-api-provider-baremetal/pkg/apis	[no test files]
?   	github.com/metalkube/cluster-api-provider-baremetal/pkg/apis/baremetal	[no test files]
2019/03/04 10:16:23 failed to start the controlplane. retried 5 times: fork/exec /usr/local/kubebuilder/bin/etcd: no such file or directory
FAIL	github.com/metalkube/cluster-api-provider-baremetal/pkg/apis/baremetal/v1alpha1	0.011s
?   	github.com/metalkube/cluster-api-provider-baremetal/pkg/cloud/baremetal/actuators/machine	[no test files]
?   	github.com/metalkube/cluster-api-provider-baremetal/pkg/controller	[no test files]
?   	github.com/metalkube/cluster-api-provider-baremetal/pkg/webhook	[no test files]
?   	github.com/metalkube/cluster-api-provider-baremetal/cmd/manager	[no test files]
make: *** [Makefile:15: test] Error 1
```

russellb:
Yeah, the tests will fail unless you install kubebuilder in the right magic location.  See this line:

```
2019/03/04 10:16:23 failed to start the controlplane. retried 5 times: fork/exec /usr/local/kubebuilder/bin/etcd: no such file or directory
```

You need to download kubebuilder and unpack it under /usr/local/.

We should add this as part of dev env setup in the README, too.

vhosakot:
[...]

PR #77, opened 2019-05-22
Makefile: Add kubeubilder and kustomize checks.

russellb:
Generating manifests and running the test suite require kustomize and
kubebuilder.  Add a quick check for them to the Makefile to avoid less
obvious errors later if they are missing.

Closes issue #14.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ test: testprereqs generate fmt vet unit
 .PHONY: testprereqs
 testprereqs:
 	@if [ ! -d /usr/local/kubebuilder ] ; then echo "kubebuilder not found.  See docs/dev/setup.md" && exit 1 ; fi
-	@if ! which kustomize >/dev/null 2>&1 ; then echo "kustomize not found.  See docs/dev/setup.md" && exit 1 ; fi
+	@if ! which kustomize >/dev/null 2>&1 ; then echo "Running 'go get sigs.k8s.io/kustomize'" && go get sigs.k8s.io/kustomize ; fi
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -11,7 +11,12 @@ build:
 all: test manager
 
 # Run tests
-test: generate fmt vet unit
+test: testprereqs generate fmt vet unit
+
+.PHONY: testprereqs
+testprereqs:
+	@if [ ! -d /usr/local/kubebuilder ] ; then echo "kubebuilder not found.  See docs/dev/setup.md" && exit 1 ; fi
+	@if ! which kustomize >/dev/null 2>&1 ; then echo "kustomize not found.  See docs/dev/setup.md" && exit 1 ; fi
```
dhellmann: We could "go get" kustomize, couldn't we?

russellb: We could.

Right now `docs/dev/setup.md` points you to `tools/install_kustomize.sh` which we had because we required an older release of kustomize.  That's not true anymore so it could be all dropped in favor of `go get`.

russellb: pushed another patch which does this

dhellmann:
Based on the failure, it looks like $PWD for the job in travis isn't the top of the repo?

russellb:
> Based on the failure, it looks like $PWD for the job in travis isn't the top of the repo?

I had removed that file, but missed that it was used in the travis-ci config

kanwar-saad:
Can one of the admins verify this patch?

PR #79, opened 2019-05-28
Do a reverse lookup on IPs

bcrochet:
For every IP found, do a reverse lookup. If it succeeds, add it as a hostname.

### DIFF
--- a/pkg/cloud/baremetal/actuators/machine/actuator.go
+++ b/pkg/cloud/baremetal/actuators/machine/actuator.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"log"
 	"math/rand"
+	"net"
 	"time"
 
 	bmh "github.com/metal3-io/baremetal-operator/pkg/apis/metal3/v1alpha1"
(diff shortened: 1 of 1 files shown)


pkg/cloud/baremetal/actuators/machine/actuator.go
```diff
@@ -428,6 +429,19 @@ func (a *Actuator) nodeAddresses(host *bmh.BareMetalHost) ([]corev1.NodeAddress,
 			Address: nic.IP,
 		}
 		addrs = append(addrs, address)
+
+		names, err := net.LookupAddr(nic.IP)
+		if err != nil {
+			log.Printf("Bad lookup - %v", err)
+			continue
+		}
+		for _, name := range names {
```
mhrivnak: I suspect it's not valid to include more than one hostname. Maybe in a case where we get more than one result, we should not add them and instead log that multiple results were found?

pkg/cloud/baremetal/actuators/machine/actuator.go
```diff
@@ -428,6 +429,19 @@ func (a *Actuator) nodeAddresses(host *bmh.BareMetalHost) ([]corev1.NodeAddress,
...
+			nameAddress := corev1.NodeAddress{
+				Type:    "Hostname",
```
mhrivnak: It would be better to reference the designated constant, `corev1.NodeHostName`.

pkg/cloud/baremetal/actuators/machine/actuator.go
bcrochet: Good suggestion. I'll default to the first found, and log future ones found.

pkg/cloud/baremetal/actuators/machine/actuator.go
bcrochet: Thanks.

bcrochet:
Closing in favor of getting hostname from IPA.

bcrochet:
Going a different direction. Using hostname from IPA and DHCP.

PR #86, opened 2019-06-17
Delay Machine deletion until BaremetalHost deprovisioned

### LINKED ISSUE #84: actuator needs to wait for host to be deprovisioned before returning from Delete()
In order to correctly fence hosts, we need to ensure that the Machine objects are not deleted until the host has been wiped, powered off, or both -- so that it is no longer trying to be a node.

It should be able to determine that the host is wiped by watching for its Status.Provisioning.Image field to have an empty image.

Today, the Machine controller appears to support returning a special type of error to indicate that a Machine needs to be requeued [1]. That works differently in the OpenShift version of the controller code [2], though, so we either need to update the copy of the cluster-api code used there or do something else.

[1] https://github.com/metal3-io/cluster-api-provider-baremetal/blob/master/vendor/sigs.k8s.io/cluster-api/pkg/controller/machine/controller.go#L172
[2] https://github.com/openshift/cluster-api-provider-baremetal/blob/master/vendor/github.com/openshift/cluster-api/pkg/controller/machine/controller.go#L206

zaneb:
In order to ensure that fencing works correctly, we can't allow the
Machine object to be deleted while there is still a chance that the
BaremetalHost is still powered on and running. Delay successful
completion of the deletion until the actuator has verified that the Host
cannot be online.

Closes #84

### DIFF
--- a/pkg/cloud/baremetal/actuators/machine/actuator.go
+++ b/pkg/cloud/baremetal/actuators/machine/actuator.go
@@ -143,24 +143,22 @@ func (a *Actuator) Delete(ctx context.Context, cluster *machinev1.Cluster, machi
 	if err != nil {
 		return err
 	}
-	if host != nil {
-		if host.Spec.MachineRef != nil {
-			// don't remove the MachineRef if it references some other machine
-			if host.Spec.MachineRef.Name == machine.Name {
-				host.Spec.MachineRef = nil
-				host.Spec.Image = nil
-				host.Spec.Online = false
-				host.Spec.UserData = nil
-				err = a.client.Update(ctx, host)
-				if err != nil && !errors.IsNotFound(err) {
-					return err
-				}
-				return &clustererror.RequeueAfterError{}
-			} else {
-				log.Printf("host associated with %v, not machine %v.",
-					host.Spec.MachineRef.Name, machine.Name)
-				return nil
+	if host != nil && host.Spec.MachineRef != nil {
+		// don't remove the MachineRef if it references some other machine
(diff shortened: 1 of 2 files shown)


pkg/cloud/baremetal/actuators/machine/actuator.go
```diff
@@ -143,18 +143,32 @@ func (a *Actuator) Delete(ctx context.Context, cluster *machinev1.Cluster, machi
...
-			host.Spec.Online = false
-			host.Spec.UserData = nil
-			err = a.client.Update(ctx, host)
-			if err != nil && !errors.IsNotFound(err) {
-				return err
+	if host != nil {
+		if host.Spec.MachineRef != nil {
+			// don't remove the MachineRef if it references some other machine
+			if host.Spec.MachineRef.Name == machine.Name {
+				host.Spec.MachineRef = nil
```
mhrivnak: In order to make this work with #90, we'd have to leave the MachineRef in place until we're ready to `return nil`.

zaneb: Hmm, yes that's true. This will still work eventually as we'll keep polling the state every 30s, but if we didn't clear the MachineRef then #90 would notify us immediately once the Host is deprovisioned so we wouldn't introduce any extra latency.

mhrivnak: FWIW I think treating the data that way would also match the intent. We want that host to remain associated with the machine until we're completely sure it is no longer functioning as a Node. If there's any chance it might still be participating in the cluster, we should preserve the association on both sides.

Issue #87, opened 2019-06-19
Machine controller does not see if a chosen BareMetalHost is deleted


russellb:
cluster-api-provider-baremetal was originally coded to assume that a BareMetalHost could not be deleted if it was associated with a Machine.  There's a bug open to prevent this case in the baremetal-operator: https://github.com/metal3-io/baremetal-operator/issues/34

The BareMetalHost spec is changing this reference from MachineRef to a more generic field, so it really shouldn't make Machine related behavior based on this field.  Instead, I think we should just ensure we reconcile Machines on BareMetalHost changes to ensure we catch a Machine that no longer has its underlying BareMetalHost.

russellb:
@mhrivnak Can you add some info on the past investigation you did around getting cluster-api-provider-baremetal to watch BareMetalHost objects?

mhrivnak:
Here is my PoC: https://github.com/mhrivnak/cluster-api-provider-baremetal/commit/f0caa5f7186a3be48407e13e8ead325d052ac41e

It's kind of a silly approach, but since the cluster API actuator pattern doesn't give us a formal way to access the `Controller`, the only way I could think of to do so was to make a fake `Manager` that wraps the real one so we could intercept the `Controller` as it's being added.

The newly-desired deletion behavior is probably a good reason to bring that in. I can dust it off and make a PR in the next day or so unless anyone has a better idea.

FWIW I did get a user story accepted in the cluster-api that covers this, so hopefully future iterations of the cluster-api will give us a more direct way to accomplish this.

dhellmann:
I wonder if just making the host object an owner of the machine object would cause a cascading delete and do what we want? If not that, then I wouldn't be opposed to adding some more generic delete protection behavior, like a finalizer on the host that the machine controller removes or a flag on the host that tells the BMO to delete its consumer when it is deleted. Anyway, I think we have a few options available.

russellb:
Good point, I'll adjust the title to be a bit more generic and not suggest a specific solution

russellb:
I believe this is resolved via #90

PR #90, opened 2019-06-21
controller watches BareMetalHosts and enqueues corresponding Machines

mhrivnak:
If a BareMetalHost that is associated with a Machine changes, the Machine
should be reconciled. This can enable (not yet implemented) logic such as
waiting for a host to deprovision while a Machine is being deleted.


pkg/manager/wrapper/mapper.go
```diff
@@ -0,0 +1,28 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+type mapper struct{}
+
+// Map will return a reconcile request for a Machine if the event is for a
+// BareMetalHost and that BareMetalHost references a Machine.
+func (m *mapper) Map(obj handler.MapObject) []reconcile.Request {
+	if host, ok := obj.Object.(*bmh.BareMetalHost); ok {
+		if host.Spec.MachineRef != nil {
```
zaneb: We're in the process of renaming this - there's no PR in this repo to do so yet, but it's something to keep in mind when we come to merge stuff.

mhrivnak:
> Just to satisfy my own curiosity: what is the reason for not using `SetControllerReference()`/`Watch()` to configure this? Is it because we're only an actuator and don't have access to the controller itself to set a Watch?

I don't 100% follow what you're asking, but maybe this answers it:

I don't think we want to establish an owner relationship where a Machine owns a BareMetalHost. If we did that, then we could indeed just Watch with `handler.EnqueueRequestForOwner`. But the main reason to establish that kind of ownership relationship would be so that Machine deletion would cause deletion of the owned BareMetalHost, which is not a behavior we want.

If anything, we might want to have the host own the machine, as suggested in #87.

> 
> IMHO it might pay for us to make a habit of doing dep updates in a separate commit from substantive changes.

Oops, agreed, I'll fix that.

> 
> BTW the waiting for deprovision thing is implemented in #86; this looks like it will work nicely with that.

I took a quick look at it earlier and it looks good. I think it'll be easy to fit with this.

russellb:
clever, nice work @mhrivnak

PR #91, opened 2019-06-25
Set the hostname from HardwareDetails

bcrochet:
We now have the Hostname from HardwareDetails. This adds it to the  MachineStatus.


pkg/cloud/baremetal/actuators/machine/actuator.go
```diff
@@ -464,5 +464,10 @@ func (a *Actuator) nodeAddresses(host *bmh.BareMetalHost) ([]corev1.NodeAddress,
 		addrs = append(addrs, address)
 	}
 
+	addrs = append(addrs, corev1.NodeAddress{
+		Type: corev1.NodeHostName,
+		Address: host.Status.HardwareDetails.Hostname,
```
mhrivnak: Is this guaranteed to always be populated? Or should we check first to ensure it has a value?

russellb: +1 to making sure it has a value first

PR #92, opened 2019-06-25
switch to using ConsumerRef instead of MachineRef

dhellmann:


pkg/cloud/baremetal/actuators/machine/actuator_test.go
```diff
@@ -34,7 +34,7 @@ func TestChooseHost(t *testing.T) {
 			Namespace: "myns",
 		},
 		Spec: bmh.BareMetalHostSpec{
-			MachineRef: &corev1.ObjectReference{
+			ConsumerRef: &corev1.ObjectReference{
```
russellb: In the code you added 'Kind: "Machine"' here, but not throughout the test code.  Should it be added to the tests as well?

dhellmann: Done

dhellmann:
> This is missing a change to the new reference to MachineRef added in #90 (in `pkg/manager/wrapper/mapper.go`).

Fixed, thanks for pointing that out.

> It also conflicts with #86, but I'm happy to update that if this merges first. (I guess updating the vendored BMO will also conflict with #91... it feels like we ought to have a better strategy for doing updates.)

Let's see about getting #86 in first. I can rebase and update this one.

> Other than that this LGTM with the stipulation that we're not expecting things to work normally after an upgrade here, since they definitely will not. Maybe we should document that in the commit message.

Yeah, one of the reasons I'm trying to get these API changes in now is we don't support upgrades, yet.

pkg/cloud/baremetal/actuators/machine/actuator.go
```diff
@@ -308,8 +308,8 @@ func (a *Actuator) chooseHost(ctx context.Context, machine *machinev1.Machine,
 			} else {
 				log.Printf("Host '%s' did not match hostSelector for Machine '%s'", host.Name, machine.Name)
 			}
-		} else if host.Spec.MachineRef != nil && host.Spec.MachineRef.Name == machine.Name && host.Spec.MachineRef.Namespace == machine.Namespace {
-			log.Printf("found host %s with existing MachineRef", host.Name)
+		} else if host.Spec.ConsumerRef != nil && host.Spec.ConsumerRef.Name == machine.Name && host.Spec.ConsumerRef.Namespace == machine.Namespace {
```
mhrivnak: This should probably verify the correct GVK. It feels a bit far-fetched that you'd encounter a BMH with a reference to the name namespace and name, but some other object type. But if we're opening up the ConsumerRef field to be used by other objects, the safest thing to do would be to check the GVK.

In cases where we already have reason to think the host is related to a Machine, such as when we're finding it based on an annotation that's on a Machine, there's a stronger argument for assuming the GVK. But even then I could see value in checking.

WDYT?

pkg/manager/wrapper/mapper.go
```diff
@@ -13,12 +13,12 @@ type mapper struct{}
 // BareMetalHost and that BareMetalHost references a Machine.
 func (m *mapper) Map(obj handler.MapObject) []reconcile.Request {
 	if host, ok := obj.Object.(*bmh.BareMetalHost); ok {
-		if host.Spec.MachineRef != nil {
+		if host.Spec.ConsumerRef != nil {
```
mhrivnak: In this case, I think we should definitely check the GVK. It would be a shame to be generating events for Machines that don't exist, in a case where some hosts have a ConsumerRef that references some other resource type.

pkg/cloud/baremetal/actuators/machine/actuator.go
dhellmann: I don't know what a GVK is or how to check it. Are you just saying I need to check all of the fields in the object reference? Can those structures be compared directly, or do I have to list them all out individually or use some sort of recursive comparison function?

pkg/cloud/baremetal/actuators/machine/actuator.go
```diff
@@ -331,7 +331,8 @@ func (a *Actuator) chooseHost(ctx context.Context, machine *machinev1.Machine,
 func (a *Actuator) setHostSpec(ctx context.Context, host *bmh.BareMetalHost, machine *machinev1.Machine,
 	config *bmv1alpha1.BareMetalMachineProviderSpec) error {
 
-	host.Spec.MachineRef = &corev1.ObjectReference{
+	host.Spec.ConsumerRef = &corev1.ObjectReference{
+		Kind:      "Machine",
```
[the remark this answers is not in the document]
dhellmann: I added checks for the APIVersion and Kind fields, and updated the code that sets the reference to include those values.

pkg/cloud/baremetal/actuators/machine/actuator.go
mhrivnak: GVK is shorthand for Group Version Kind, for example "metal3.io", "v1alpha1", and "BareMetalHost". Group and Version are sometimes combined and simply called "APIVersion", such as "metal3.io/v1alpha1".

We just need to make sure as necessary that the `ConsumerRef` is referencing an object of the type this controller cares about., and the GVK is what fully identified a resource type. I don't know how much help is available from a library with the comparison, but you might find something useful [here](https://godoc.org/k8s.io/apimachinery/pkg/runtime/schema#GroupVersionKind).

pkg/manager/wrapper/mapper.go
```diff
@@ -13,12 +13,12 @@ type mapper struct{}
 // BareMetalHost and that BareMetalHost references a Machine.
 func (m *mapper) Map(obj handler.MapObject) []reconcile.Request {
 	if host, ok := obj.Object.(*bmh.BareMetalHost); ok {
-		if host.Spec.MachineRef != nil {
+		if host.Spec.ConsumerRef != nil && host.Spec.ConsumerRef.Kind == "Machine" {
```
mhrivnak: It would be good to check the APIVersion here also, so we have a complete type reference.

pkg/cloud/baremetal/actuators/machine/actuator_test.go
```diff
@@ -34,9 +34,11 @@ func TestChooseHost(t *testing.T) {
...
 		},
 		Spec: bmh.BareMetalHostSpec{
-			MachineRef: &corev1.ObjectReference{
-				Name:      "someothermachine",
-				Namespace: "myns",
+			ConsumerRef: &corev1.ObjectReference{
+				Name:       "someothermachine",
+				Namespace:  "myns",
+				Kind:       "Machine",
+				APIVersion: "v1alpha1",
```
mhrivnak: Just want to clarify one point, which is confusing about these field names. "APIVersion" should contain both the group and version, for example `metal3.io/v1alpha1`. It's hard to know that except by seeing examples [like this](https://github.com/kubernetes/kubernetes/blob/7f23a743e8c23ac6489340bbb34fa6f1d392db9d/staging/src/k8s.io/client-go/tools/reference/ref.go#L77).

Not including a group [implies](https://kubernetes.io/docs/concepts/overview/kubernetes-api/#api-groups) the k8s core group, for example the things found under `/api/v1/`.

In test data it doesn't make a whole lot of difference of course, but it's probably better to include the group as it will be included in real use.

dhellmann: I started to do that and then discovered that the group and version are different for Machine upstream vs. OpenShift. So I decided to just leave it as it is for the tests.

pkg/manager/wrapper/mapper.go
dhellmann: The APIversion for OpenShift is "machine.openshift.io/v1beta1" but upstream is on v1alpha1 and won't be using machine.openshift.io as the group. Is there some way to discover the group and version at runtime without having a Machine instance?

mhrivnak: I think it would be reasonable to create an empty Machine instance and grab the values from it. The nice thing about managing the openshift vs upstream maintenance is that we can usually make the change just in import statements, as long as the imports use the same alias.

dhellmann:
I found a place where we can get the group and version dynamically, so I've updated the code to use that.

dhellmann:
I have rebased this after merging #86 and added some changes to deal with the machine references in that PR. This is ready for a re-review.

PR #97, opened 2019-07-09
Allow deletion of externally provisioned Hosts

zaneb:
If a Host is in an "externally provisioned" state, then we have no control over the Host's provisioning state. Therefore when deleting a Machine with an externally provisioned Host, we should wait only until
the Host has been powered off rather than for it to be deprovisioned (which will never happen).

### DIFF
--- a/pkg/cloud/baremetal/actuators/machine/actuator.go
+++ b/pkg/cloud/baremetal/actuators/machine/actuator.go
@@ -160,12 +160,22 @@ func (a *Actuator) Delete(ctx context.Context, cluster *machinev1.Cluster, machi
 			}
 			return &clustererror.RequeueAfterError{}
 		}
+
+		waiting := true
 		switch host.Status.Provisioning.State {
-		default:
-			return &clustererror.RequeueAfterError{RequeueAfter: requeueAfter}
 		case bmh.StateRegistrationError, bmh.StateRegistering,
 			bmh.StateMatchProfile, bmh.StateInspecting,
 			bmh.StateReady, bmh.StateValidationError:
+			// Host is not provisioned
+			waiting = false
+		case bmh.StateExternallyProvisioned:
+			// We have no control over provisioning, so just wait until the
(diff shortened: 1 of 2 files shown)


pkg/cloud/baremetal/actuators/machine/actuator.go
```diff
@@ -160,12 +160,22 @@ func (a *Actuator) Delete(ctx context.Context, cluster *machinev1.Cluster, machi
...
-			return &clustererror.RequeueAfterError{RequeueAfter: requeueAfter}
 		case bmh.StateRegistrationError, bmh.StateRegistering,
 			bmh.StateMatchProfile, bmh.StateInspecting,
 			bmh.StateReady, bmh.StateValidationError:
+			// Host is not provisioned
+			waiting = false
+		case bmh.StateExternallyProvisioned:
+			// We have no control over provisioning, so just wait until the
+			// host is powered off
+			waiting = host.Status.PoweredOn
```
russellb: What will ensure the host is powered off?  AFAIK, the baremetal-operator keeps all hosts powered on (with online: true), even when deprovisioned.  It seems like this is still leaving a case where deletion could be stuck forever.

russellb:
Please test this with metal3-dev-env if you have not done so already

pkg/cloud/baremetal/actuators/machine/actuator.go
zaneb: We turn the power off on line 155. The operator still does power management of externally provisioned hosts: https://github.com/metal3-io/baremetal-operator/blob/master/pkg/controller/baremetalhost/baremetalhost_controller.go#L322

zaneb:
I tested this with metal3-openshift/dev-scripts and the Machine did delete.

russellb:
this works for me too, thanks

jan-est:
This works for me. Now it deprovisioning working as it should. Thanks a lot.

PR #99, opened 2019-07-22
kustomization.yaml: Use patchesStrategicMerge.

russellb:
Change "patches" to "patchesStrategicMerge".  The use of "patches" has
been deprecated since v1.0.9:

https://github.com/kubernetes-sigs/kustomize/blob/v1.0.9/pkg/types/kustomization.go#L129

With kustomize v3, this fails with a message similar to the one seen
in this issue:

https://github.com/kubernetes-sigs/kustomize/issues/1373

This change should not result in any change in behavior, but makes
this kustomization.yaml no longer use a deprecated field.


russellb:
I did a big vendor updated to get the same fix in our vendored copy of cluster-api.  To get the vendor update to work properly, i ended up having to just update everything.

mhrivnak:
Looks reasonable to me. Just need to solve that dependency problem.

russellb:
> Looks reasonable to me. Just need to solve that dependency problem.

forgot to `git add` new files in `vendor/`.  CI passes now

PR #102, opened 2019-08-15
ensure all hosts that are in use have an image reference

dhellmann:
Externally provisioned hosts need the image as well, so always set it
unless it is already set.


dhellmann:
This is part 2 of the work started in https://github.com/metal3-io/baremetal-operator/pull/278

PR #114, opened 2019-10-09
Add hack scripts to run tests in containers.

russellb:
These scripts will make it easy to run each of these tests in the same
container that is used to run it in CI.


russellb:
This is a prerequisite for https://github.com/metal3-io/project-infra/issues/18

PR #117, opened 2019-10-09
Fix broken deployment example and update kustomize

maelk:
The deployment example for the cluster was broken when modifying the API, fixing it and updating kustomize


maelk:
I'll add the missing scripts to run the tests

maelk:
I will fix the linting errors

russellb:
Instead of the new commit to add the scripts you could have merged master into this branch. We will have to deal with any branch conflicts eventually either way.

PR #118, opened 2019-10-10
[V1alpha2] Refactor the Makefile targets

maelk:
now deploy examples by running for example
make deploy; make deploy-examples

The deployment of the CRDs and controllers is not included in the
examples anymore

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -182,27 +182,14 @@ generate-manifests: $(CONTROLLER_GEN) ## Generate manifests e.g. CRD, RBAC etc.
 		output:crd:dir=$(CRD_ROOT) \
 		output:webhook:dir=$(WEBHOOK_ROOT) \
 		webhook
-	# baremetal-operator v1 alpha1 is not compliant with kubernetes
-	# $(CONTROLLER_GEN) \
-	#	paths=github.com/metal3-io/baremetal-operator/pkg/apis/... \
-	#	crd:trivialVersions=true \
-	#	output:crd:dir=$(METAL3_CRD_ROOT) \
-	#	output:none
 	$(CONTROLLER_GEN) \
 		paths=./controllers/... \
 		output:rbac:dir=$(RBAC_ROOT) \
 		rbac:roleName=manager-role
 
 .PHONY: generate-examples
-generate-examples: clean-examples ## Generate examples configurations to run a cluster.
+generate-examples: manifests clean-examples ## Generate examples configurations to run a cluster.
 	./examples/generate.sh
-
-apply-examples: generate-examples
-	kubectl apply -f ./examples/_out/provider-components.yaml
-	kubectl apply -f ./examples/_out/metal3plane.yaml
-	kubectl apply -f ./examples/_out/cluster.yaml
-	kubectl apply -f ./examples/_out/machinedeployment.yaml
-	kubectl apply -f ./examples/_out/controlplane.yaml
 
 ## --------------------------------------
 ## Docker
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -266,26 +253,33 @@ manifests: generate-manifests
...
 
 # Deploy controller in the configured Kubernetes cluster in ~/.kube/config
-deploy: manifests deploy-capi
+deploy: manifests
 	kubectl apply -f examples/provider-components/provider-components-baremetal.yaml
 
 # Deploy CAPI controllers in the configured Kubernetes cluster in ~/.kube/config
 deploy-capi: manifests
 	kubectl apply -f examples/provider-components/provider-components-cluster-api.yaml
 	kubectl apply -f examples/provider-components/provider-components-kubeadm.yaml
```
jbrette: This will fail if the generate.sh is modified. Those files are only here because generate.sh did create them and passed the right parameters during the kustomize build

examples/generate.sh
```diff
@@ -96,21 +91,3 @@ echo "Generated ${METAL3PLANE_GENERATED_FILE}"
...
-
-# Generate Kubeadm Bootstrap Provider components file.
-# kustomize build "github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm//config/default/?ref=master" > "${COMPONENTS_KUBEADM_GENERATED_FILE}"
-kustomize build "${SOURCE_DIR}/../../cluster-api-bootstrap-provider-kubeadm/config/default" > "${COMPONENTS_KUBEADM_GENERATED_FILE}"
-echo "Generated ${COMPONENTS_KUBEADM_GENERATED_FILE}"
-
-# Generate BAREMETAL Infrastructure Provider components file.
-kustomize build "${SOURCE_DIR}/../config/default" | envsubst > "${COMPONENTS_BAREMETAL_GENERATED_FILE}"
-echo "Generated ${COMPONENTS_BAREMETAL_GENERATED_FILE}"
-
```
jbrette: This needs to stay until we find a clean way to have a top level kustomize.yaml and not have to do that in a shell script

examples/generate.sh
```diff
@@ -96,21 +91,3 @@ echo "Generated ${METAL3PLANE_GENERATED_FILE}"
...
-# Generate a single provider components file.
-kustomize build "${SOURCE_DIR}/provider-components" | envsubst > "${PROVIDER_COMPONENTS_GENERATED_FILE}"
-echo "Generated ${PROVIDER_COMPONENTS_GENERATED_FILE}"
```
jbrette: Needs to stay.

Makefile
```diff
@@ -266,26 +253,33 @@ manifests: generate-manifests
...
 	go run ./main.go
 
 # Install CRDs into a cluster
 install: manifests
 	kubectl apply -k config/crd
 
 # Deploy controller in the configured Kubernetes cluster in ~/.kube/config
-deploy: manifests deploy-capi
+deploy: manifests
 	kubectl apply -f examples/provider-components/provider-components-baremetal.yaml
```
jbrette: This is not suppposed to work if the baremetalhosts have not been deployed definition has not been deployed.

Makefile
maelk: There is a makefile target to generate those, it does not need to be in generate.sh. My idea is to clean generate.sh so that it actually only generates the example CRs, not the CRDs and controller deployments around it

examples/generate.sh
maelk: This is done in the Makefile (manifests target)

examples/generate.sh
maelk: It is also in the Makefile

Makefile
maelk: The BMH CRDs are part of that file, so it is self-contained.

examples/generate.sh
maelk: `make deploy-capi, make-deploy` will do the same thing as `./generate.sh; kubectl apply -f _out/provider-components.yaml`

jbrette:
I think this is still wrong for two reasons:

1) baremetal is back to one of kind behavior because all the other providers implements the installation of the capi and capbk through generate.sh. kustomize build <git> is extremely limited and boggus and should not be used. This is why the other components are doing wget, curl...to deal with that bug in kustomize.

2) but mainly: You can't configure anymore the capi nor cabpk operators behavior, pullPolicy or images. Your stuck to latest version of whatever has been tagged. This is the main purpose of that line. 

```bash
kustomize build "${SOURCE_DIR}/provider-components" | envsubst > "${PROVIDER_COMPONENTS_GENERATED_FILE}"
```
because you can change the images, pullpolicy from that kustomization.yaml.

At minimum, the Makefile should invoke the kustomize build of the provider-components directory.

maelk:
Comments from @jbrette fixed.

PR #119, opened 2019-10-10
Create and use a unit test image.

russellb:
Instead of installing kubebuilder and kustomize each time the unit
tests run in a container, just pre-install those in a container image.

Also update install_kubebuilder.sh to remove the downloaded tar.gz to
reduce the size of the resulting container image.


russellb:
Oops, i shouldn't have done "/test unit", i need to update project-config to use the new image first

PR #121, opened 2019-10-10
hack/govet: Set XDG_CACHE_HOME

russellb:
When running under prow, running "go vet" was trying to write to
/.cache, which doesn't work.  Try setting an env var to move this
under /tmp.


russellb:
still need to verify that this fixes the problem

/hold

russellb:
yay, it works.

/hold cancel

PR #126, opened 2019-10-14
Add make and git command to capbm-unit Dockerfile

maelk:
Fix needed for v1alpha2 work #125 to run the unit tests properly


maelk:
Creating a v1alpha2 image and specific jobs instead

PR #127, opened 2019-10-15
Enable leader election

### LINKED ISSUE #113: LeaderElectionID should be unique
This issue was reported in other providers - Running several Cluster API pods (e.g. CAPI + CABPK + an infra provider) in the same namespace results in leader election errors. 
More details here - kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm#271

zaneb:
Fixes #113

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -63,6 +63,8 @@ func main() {
 	// Setup a Manager
 	opts := manager.Options{
 		MetricsBindAddress: *metricsAddr,
+		LeaderElection:     true,
+		LeaderElectionID:   "controller-leader-election-capbm",
 	}
 	if *watchNamespace != "" {
 		opts.Namespace = *watchNamespace


maelk:
Here is what we did in capbm v1alpha2 related to this : https://github.com/metal3-io/cluster-api-provider-baremetal/blob/v1alpha2/main.go#L62 and https://github.com/metal3-io/cluster-api-provider-baremetal/blob/v1alpha2/main.go#L73

mhrivnak:
Another option is to detect if you're running in a cluster and only use leader election if so. That's how operator-sdk's leader election works by default: https://github.com/operator-framework/operator-sdk/blob/418b603/pkg/leader/leader.go#L48-L55

maelk:
@zaneb This should now be for the v1alpha1 branch, not master (that is v1alpha2)

maelk:
Closing this issue since it is done in master, and the v1alpha1 branch will be deleted

PR #146, opened 2019-10-23
Move BMH CRDs from config/crd to examples/metal3crds

maelk:
for testing without bmo, requires running make deploy-bmo-cr before
deploying
Their current place creates a bug if not synchronized with Baremetal Operator


kashifest:
lgtm but lets wait for @jan-est's comment

PR #148, opened 2019-10-24
Remove unneeded CAPD functions

maelk:
Kubeadmreset is not supported, will be part of CAPI control plane
management, and execBoostrap was a replacement for cloud-init in CAPD

### DIFF
--- a/controllers/baremetalmachine_controller.go
+++ b/controllers/baremetalmachine_controller.go
@@ -175,20 +175,6 @@ func (r *BareMetalMachineReconciler) reconcileNormal(ctx context.Context,
 		return ctrl.Result{}, errors.Wrap(err, "failed to create worker BareMetalMachine")
 	}
 
-	// if the machine is a control plane added, update the load balancer configuration
-	if util.IsControlPlaneMachine(machineMgr.Machine) {
-		if err := clusterMgr.UpdateConfiguration(); err != nil {
-			return ctrl.Result{}, errors.Wrap(err, "failed to update BareMetalCluster.loadbalancer configuration")
-		}
-	}
-
-	// exec bootstrap
-	// NB. this step is necessary to mimic the behaviour of cloud-init that is embedded in the base images
-	// for other cloud providers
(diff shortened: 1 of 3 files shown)


controllers/baremetalmachine_controller.go
```diff
@@ -200,13 +186,6 @@ func (r *BareMetalMachineReconciler) reconcileNormal(ctx context.Context,
 
 func (r *BareMetalMachineReconciler) reconcileDelete(ctx context.Context,
 	machineMgr *baremetal.MachineManager, clusterMgr *baremetal.ClusterManager) (ctrl.Result, error) {
-	// if the deleted machine is a control-plane node, exec kubeadm reset so the etcd member hosted
-	// on the machine gets removed in a controlled way
-	if util.IsControlPlaneMachine(machineMgr.Machine) {
-		if err := machineMgr.KubeadmReset(); err != nil {
-			return ctrl.Result{}, errors.Wrap(err, "failed to execute kubeadm reset")
-		}
-	}
```
Xenwar: Do we plan to add 'decommissioning' of nodes ?
If so, we need to leave a skeleton

controllers/baremetalmachine_controller.go
```diff
@@ -218,13 +197,6 @@ func (r *BareMetalMachineReconciler) reconcileDelete(ctx context.Context,
 		}
 	}
 
-	// if the deleted machine is a control-plane node, remove it from the load balancer configuration;
```
Xenwar: If we are managing(configuring) the LB from CAPBM, we might need to keep the skeleton.

controllers/baremetalmachine_controller.go
maelk: This will be part of CAPI v1alpha3 (check the control plane management CAEP). It will not be done by the infrastructure provider

controllers/baremetalmachine_controller.go
maelk: let's add it back when we have an actual solution for the LB

PR #153, opened 2019-10-31
[V1alpha2] Modify baremetalmachine workflow

maelk:
Wait for the baremetalhost to be provisioned before setting
the providerID and ready flag on the baremetalmachine


baremetal/machine_manager.go
```diff
@@ -145,22 +157,21 @@ func (mgr *MachineManager) Close() error {
 }
 
 // Create creates a machine and is invoked by the Machine Controller
```
kashifest: Change the description accordingly

PR #166, opened 2019-11-13
Remove setting error status on non-existing BMMachine

Xenwar:


controllers/baremetalmachine_controller.go
```diff
@@ -255,6 +288,12 @@ func (r *BareMetalMachineReconciler) BareMetalClusterToBareMetalMachines(o handl
 	case apierrors.IsNotFound(err) || cluster == nil:
 		return result
 	case err != nil:
+		// DELETE_ME before merging
+		// whatever the reason, owning cluster is not found. either missing label, owner ref...
+		// we blame config error
+		c.Status.ErrorMessage = pointer.StringPtr("failed to get owning cluster")
+		v := capierrors.InvalidConfigurationClusterError
+		c.Status.ErrorReason = &v
```
Jaakko-Os: why not write a function that would take the 'string' and client/machine/bmmachine as input?
code could be something like:
if "bmmachine" {
capbmMachine.Status.ErrorMessage = pointer.StringPtr("BareMetalMachine's owner Machine is missing cluster label or cluster does not exist")
v := capierrors.InvalidConfigurationMachineError
capbmMachine.Status.ErrorReason = &v
}
and so on.

Xenwar: +1 for this. Will try to do that.

Xenwar: Done

Jaakko-Os: /lgtm

Jaakko-Os: Good work!

controllers/baremetalcluster_controller.go
```diff
@@ -114,6 +122,9 @@ func reconcileNormal(ctx context.Context, clusterMgr *baremetal.ClusterManager)
 
 	// Set APIEndpoints so the Cluster API Cluster Controller can pull it
 	if err := clusterMgr.UpdateClusterStatus(); err != nil {
+		//er := errors.New("failed to get ip for the API endpoint")
+		//setErrorBMCluster(clusterMgr.BareMetalCluster, er, capierrors.CreateMachineError, capierrors.InvalidConfigurationClusterError)
+
```
maelk: Don't add commented code without giving the reason. It's better to either not put it at all or not comment it if it is working

controllers/baremetalcluster_controller.go
```diff
@@ -141,7 +153,10 @@ func (r *BareMetalClusterReconciler) reconcileDelete(ctx context.Context,
 	}
 
 	if err := clusterMgr.Delete(); err != nil {
-		return ctrl.Result{}, errors.Wrap(err, "failed to delete load balancer")
+		//er := errors.New("failed to delete BareMetalCluster")
+		//setErrorBMCluster(clusterMgr.BareMetalCluster, er, capierrors.CreateMachineError, capierrors.DeleteClusterError)
```
maelk: same here

controllers/baremetalmachine_controller.go
```diff
@@ -63,18 +66,27 @@ func (r *BareMetalMachineReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result,
 
 	// Fetch the BareMetalMachine instance.
 	capbmMachine := &capbm.BareMetalMachine{}
+
 	if err := r.Client.Get(ctx, req.NamespacedName, capbmMachine); err != nil {
 		if apierrors.IsNotFound(err) {
 			return ctrl.Result{}, nil
 		}
+		er := errors.Wrapf(err, fmt.Sprintf("%#v", apierrors.ReasonForError(err)))
+		setErrorBMMachine(capbmMachine, err, capierrors.CreateMachineError)
```
maelk: if we are here, it is because we failed to fetch the baremetalmachine, so it is useless to set an error in it as it is a blank object, and we won't act on it

controllers/baremetalmachine_controller.go
```diff
@@ -86,8 +98,11 @@ func (r *BareMetalMachineReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result,
 	// Fetch the Cluster.
 	cluster, err := util.GetClusterFromMetadata(ctx, r.Client, capiMachine.ObjectMeta)
 	if err != nil {
+		er := errors.Wrapf(err, "BareMetalMachine's owner Machine is missing cluster label or cluster does not exist")
 		machineLog.Info("BareMetalMachine's owner Machine is missing cluster label or cluster does not exist")
-		return ctrl.Result{}, errors.Wrapf(err, "BareMetalMachine's owner Machine is missing cluster label or cluster does not exist")
+		setErrorBMMachine(capbmMachine, er, capierrors.InvalidConfigurationMachineError)
+
+		return ctrl.Result{}, errors.Wrapf(err, "Barer label or cluster does not exist")
```
maelk: Barer ?

controllers/baremetalmachine_controller.go
```diff
@@ -293,3 +322,21 @@ func (r *BareMetalMachineReconciler) BareMetalHostToBareMetalMachines(obj handle
...
+
+}
+
+// clearError removes the ErrorMessage from the baremetalmachine's Status if set.
+func clearErrorBMMachine(bmm *capbm.BareMetalMachine) {
+
+	if bmm.Status.ErrorMessage != nil || bmm.Status.ErrorReason != nil {
+		bmm.Status.ErrorMessage = nil
+		bmm.Status.ErrorReason = nil
+	}
```
maelk: This is already defined in machine_manager.go please use those rather than creating the same functions here

controllers/baremetalcluster_controller.go
Xenwar: fixed in a separate PR.

controllers/baremetalcluster_controller.go
Xenwar: fixed in a separate PR.

controllers/baremetalmachine_controller.go
Xenwar: removed

controllers/baremetalmachine_controller.go
Xenwar: fixed

controllers/baremetalmachine_controller.go
Xenwar: We cannot use those as they are specific to a resource. 
Tried passing interface to the setError and clearError functions, to no avail. 
The work around was to declare such methods per resource.

PR #167, opened 2019-11-13
Add new error type APIEndPointError

kashifest:
Add APIEndPointError type and refactor baremetalcluster_types.isValid() function


api/v1alpha2/baremetalcluster_types.go
```diff
@@ -38,16 +38,20 @@ type BareMetalClusterSpec struct {
 // IsValid returns an error if the object is not valid, otherwise nil. The
 // string representation of the error is suitable for human consumption.
 func (s *BareMetalClusterSpec) IsValid() error {
-	missing := []string{}
 	if s.APIEndpoint == "" {
-		missing = append(missing, "APIEndpoint")
-	}
-	if len(missing) > 0 {
-		return fmt.Errorf("Missing fields from Spec: %v", missing)
+		return &APIEndPointError{"Missing fields from Spec: APIEndPoint"}
```
maelk: I would keep this, it gives us the possibility to extend the checks later on if we add fields

api/v1alpha2/apiendpoint_error.go
```diff
@@ -0,0 +1,34 @@
+/*
```
maelk: maybe this does not need a new file ?

baremetal/cluster_manager.go
```diff
@@ -112,22 +112,10 @@ func (s *ClusterManager) APIEndpoints() ([]capbm.APIEndpoint, error) {
...
-
 	if p == "" {
 		p = APIEndpointPort
 	}
-	port, err := strconv.Atoi(p)
-
-	if err != nil {
-		s.Log.Error(err, "Invalid Port")
-		return nil, err
-	}
```
maelk: Keeping the error handling here could be safer, for example if the check is for a reason or another removed from isvalid function

PR #175, opened 2019-11-20
Sync and verify error status and message on BMMachine object

Xenwar:
Sets and Verifies BMMachine Error status

### DIFF
--- a/controllers/baremetalmachine_controller.go
+++ b/controllers/baremetalmachine_controller.go
@@ -32,6 +32,7 @@ import (
 	capi "sigs.k8s.io/cluster-api/api/v1alpha2"
 	capierrors "sigs.k8s.io/cluster-api/errors"
 	"sigs.k8s.io/cluster-api/util"
+	"sigs.k8s.io/cluster-api/util/patch"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
@@ -73,6 +74,17 @@ func (r *BareMetalMachineReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result,
 		}
 		return ctrl.Result{}, err
 	}
+	helper, err := patch.NewHelper(capbmMachine, r.Client)
+	if err != nil {
+		return ctrl.Result{}, errors.Wrap(err, "failed to init patch helper")
+	}
+	// Always patch capbMachine exiting this function so we can persist any BareMetalMachine changes.
+	defer func() {
+		err := helper.Patch(ctx, capbmMachine)
+		if err != nil {
+			machineLog.Info("failed to Patch capbmMachine")
+		}
+	}()
 	//clear an error if one was previously set
 	clearErrorBMMachine(capbmMachine)
 
@@ -80,9 +92,7 @@ func (r *BareMetalMachineReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result,
 	capiMachine, err := util.GetOwnerMachine(ctx, r.Client, capbmMachine.ObjectMeta)
 
 	if err != nil {
-		er := errors.Wrapf(err, "BareMetalMachine's owner Machine could not be retrieved")
-		setErrorBMMachine(capbmMachine, er, capierrors.CreateMachineError)
-		return ctrl.Result{}, er
+		return ctrl.Result{}, errors.Wrapf(err, "BareMetalMachine's owner Machine could not be retrieved")
 	}
 	if capiMachine == nil {
 		machineLog.Info("Waiting for Machine Controller to set OwnerRef on BareMetalMachine")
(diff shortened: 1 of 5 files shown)


controllers/baremetalmachine_controller.go
```diff
@@ -66,7 +67,17 @@ func (r *BareMetalMachineReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result,
 
 	// Fetch the BareMetalMachine instance.
 	capbmMachine := &capbm.BareMetalMachine{}
-
+	helper, err := patch.NewHelper(capbmMachine, r.Client)
```
maelk: The current code was doing this with machine. is this fixing copy-paste or something else ?

controllers/baremetalmachine_controller.go
```diff
@@ -94,7 +103,7 @@ func (r *BareMetalMachineReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result,
 	// Fetch the Cluster.
 	cluster, err := util.GetClusterFromMetadata(ctx, r.Client, capiMachine.ObjectMeta)
 	if err != nil {
-		er := errors.Wrapf(err, "BareMetalMachine's owner Machine is missing cluster label or cluster does not exist")
```
maelk: By doing this you are hiding the actual error, that will make debugging harder since the error won't appear in the logs

controllers/baremetalmachine_controller_test.go
```diff
@@ -79,6 +87,11 @@ var _ = Describe("Reconcile Baremetalcluster", func() {
 			} else {
 				Expect(res.Requeue).To(BeFalse())
 			}
+			if tc.ErrorReasonExpected {
+				_ = c.Get(context.TODO(), key, testbmm)
+				Expect(testbmm.Status.ErrorReason).NotTo(Equal(nil))
```
maelk: Incorrect syntax. should be NotTo(BeNil())

controllers/baremetalmachine_controller.go
Xenwar: not clear. This a new content and unique per resource.

controllers/baremetalmachine_controller_test.go
Xenwar: The syntax is correct. 
but changed as suggested.

controllers/baremetalmachine_controller.go
Xenwar: This pattern is used though the code. We need to have a convention on what style to use.

maelk: actually no, the New is only used when we do not have an error already. When we do, we just wrap

maelk:
@Xenwar please rebase

controllers/baremetalmachine_controller.go
```diff
@@ -141,13 +150,6 @@ func (r *BareMetalMachineReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result,
 		return ctrl.Result{}, errors.Wrapf(err, "failed to create helper for managing the clusterMgr")
 	}
 
-	// Always patch capbmMachine when exiting this function so we can persist any BareMetalMachine changes.
-	defer func() {
-		if err := machineMgr.Close(); err != nil && rerr == nil {
-			rerr = err
-		}
-	}()
-
```
maelk: maybe you should remove the close function from the machine manager code if it is not used anymore. The patch helper in this close function was using a machine, not a baremetalmachine (related to my first comment). What is the impact of this change ?

maelk: to be clear, I'm talking about line 80 of machine_manager.go

Xenwar: This commit is for BMMachine from which the close references are removed. 
We need to handle Machine and Clusters separately.

Jaakko-Os: No reference to Close() in machine mgr unit tests, therefore it can be removed.

Jaakko-Os: No reference to Close() in machine mgr unit tests, therefore it can be removed.

Xenwar: Got more info from @Jaakko-Os. And, close function is removed from the code.

Jaakko-Os: looks good!

controllers/baremetalmachine_controller.go
Xenwar: removed the wrapping from the `return`

controllers/baremetalmachine_controller.go
```diff
@@ -94,11 +104,11 @@ func (r *BareMetalMachineReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result,
 	// Fetch the Cluster.
 	cluster, err := util.GetClusterFromMetadata(ctx, r.Client, capiMachine.ObjectMeta)
 	if err != nil {
-		er := errors.Wrapf(err, "BareMetalMachine's owner Machine is missing cluster label or cluster does not exist")
+		er := errors.New("BareMetalMachine's owner Machine is missing cluster label or cluster does not exist")
 		machineLog.Info("BareMetalMachine's owner Machine is missing cluster label or cluster does not exist")
 		setErrorBMMachine(capbmMachine, er, capierrors.InvalidConfigurationMachineError)
 
-		return ctrl.Result{}, errors.Wrapf(err, "BareMetalMachine's owner Machine is missing label or the cluster does not exist")
```
kashifest: In my opinion, this wrapping is fine as long as it is wrapped around the actual error

PR #185, opened 2019-11-22
[v1alpha2] Use interfaces for the managers

maelk:
This enables us to mock the managers later on to have actual unit tests

### DIFF
--- a/baremetal/baremetalmachine_manager.go
+++ b/baremetal/baremetalmachine_manager.go
@@ -57,6 +57,31 @@ const (
 	requeueAfter   = time.Second * 30
 )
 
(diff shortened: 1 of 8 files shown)


baremetal/baremetalmachine_manager.go
```diff
@@ -69,7 +94,7 @@ type MachineManager struct {
 }
 
 // NewMachineManager returns a new helper for managing a cluster with a given name.
-func newMachineManager(client client.Client,
+func NewMachineManager(client client.Client,
```
Xenwar: Would this need all four resources ?. If so, update the comment accordingly.

maelk:
In #186.

PR #187, opened 2019-11-25
Refactor baremetal machine controller and manager code

maelk:
Improve the interface between controller and manager for the baremetal machine


baremetal/baremetalmachine_manager.go
```diff
@@ -202,23 +182,29 @@ func (m *MachineManager) Associate(ctx context.Context) error {
 	err := config.IsValid()
 	if err != nil {
 		// Should have been picked earlier. Do not requeue
-		m.setError(ctx, err.Error(), capierrors.InvalidConfigurationMachineError)
+		m.setError(err.Error(), capierrors.InvalidConfigurationMachineError)
```
Xenwar: Change err.Error()
It needs to be plain string

maelk: err.Error() returns a string

maelk: we will deal with coding guidelines in a further commit

baremetal/baremetalmachine_manager.go
```diff
@@ -326,8 +323,12 @@ func (m *MachineManager) Delete(ctx context.Context) error {
 			host.Spec.UserData = nil
 			err = m.client.Update(ctx, host)
 			if err != nil && !apierrors.IsNotFound(err) {
+				m.setError("Failed to delete BareMetalMachine",
```
Xenwar: comment for line 315:
m.Log.Info("host is already associated with another bare metal machine", "host",

baremetal/baremetalmachine_manager.go
```diff
@@ -632,15 +627,15 @@ func (m *MachineManager) HasAnnotation() bool {
...
 	m.BareMetalMachine.Status.ErrorMessage = &message
 	m.BareMetalMachine.Status.ErrorReason = &reason
 }
 
 // clearError removes the ErrorMessage from the machine's Status if set. Returns
 // nil if ErrorMessage was already nil. Returns a RequeueAfterError if the
 // machine was updated.
-func (m *MachineManager) clearError(ctx context.Context) {
+func (m *MachineManager) clearError() {
 	if m.BareMetalMachine.Status.ErrorMessage != nil || m.BareMetalMachine.Status.ErrorReason != nil {
```
Xenwar: would one of them be enough ?
IN case developers could not find relevant value for one of them ?

baremetal/baremetalmachine_manager.go
```diff
@@ -730,5 +728,13 @@ func (m *MachineManager) SetNodeProviderID(ctx context.Context, bmhID, providerI
 			return errors.Wrap(err, "unable to update node for baremetal host")
```
Xenwar: ... update node associated with baremetal host

baremetal/baremetalmachine_manager_test.go
```diff
@@ -173,6 +176,216 @@ var bmhStatus = &bmh.BareMetalHostStatus{
...
+		},
+		"missing providerID": {
+			BMMachine: capbm.BareMetalMachine{
+				Status: capbm.BareMetalMachineStatus{
+					Ready: true,
+				},
+			},
+			ExpectTrue: false,
+		},
+		"nothing ready": {
```
Xenwar: replace nothing with what is missing. 
eg. Missing provierID and Ready fields

baremetal/manager_factory_test.go
```diff
@@ -0,0 +1,66 @@
+/*
```
Xenwar: Review marker.
Next review starts here.

baremetal/baremetalmachine_manager.go
maelk: I don't understand this one, sorry

baremetal/requeue_error_test.go
```diff
@@ -0,0 +1,37 @@
...
+package baremetal
+
+import (
+	"testing"
+	"time"
+)
+
+func TestError(t *testing.T) {
+	err := &RequeueAfterError{30}
+	if err.Error() != "requeue in: 30ns" {
```
Xenwar: replace to a value other than 30. One may not notice the 'ns'

baremetal/requeue_error_test.go
```diff
@@ -0,0 +1,37 @@
...
+		t.Errorf("Error, expected 30, got %s", err.RequeueAfter)
+	}
+}
+
+func TestGetRequeueAfter(t *testing.T) {
+	duration, _ := time.ParseDuration("30ns")
```
Xenwar: change 30ns to some other value, eg. 40 sec

controllers/baremetalmachine_controller.go
```diff
@@ -186,36 +177,23 @@ func (r *BareMetalMachineReconciler) reconcileNormal(ctx context.Context,
 		//Associate the baremetalhost hosting the machine
 		err := machineMgr.Associate(ctx)
 		if err != nil {
-			setErrorBMMachine(machineMgr.GetBareMetalMachine(), "failed to associate the BareMetalMachine to a BaremetalHost", capierrors.CreateMachineError)
-			return ctrl.Result{}, err
+			return checkError(err, "failed to associate the BareMetalMachine to a BaremetalHost")
```
Xenwar: you need to set ErrorReason and ErrorMessage.
Do it either before calling checkError or inside it.
In the latter case, you need to send both meesage and reason

controllers/baremetalmachine_controller.go
```diff
@@ -224,22 +202,15 @@ func (r *BareMetalMachineReconciler) reconcileNormal(ctx context.Context,
...
-		}
-
-		setErrorBMMachine(machineMgr.GetBareMetalMachine(), "failed to delete BareMetalMachine", capierrors.DeleteMachineError)
-		return ctrl.Result{}, err
+		return checkError(err, "failed to delete BareMetalMachine")
 	}
 
 	// Machine is deleted so remove the finalizer.
-	machineMgr.GetBareMetalMachine().Finalizers = util.Filter(machineMgr.GetBareMetalMachine().Finalizers, capbm.MachineFinalizer)
+	machineMgr.UnsetFinalizer()
```
Xenwar: remove from where ?
It seems that there is another resource that is not found and we removing the finalizer from the machine.
If no code change is required, then update the comment

controllers/baremetalmachine_controller.go
maelk: the error reason and messages are set in the Associate function, doing it again here would overwrite

baremetal/baremetalmachine_manager.go
Xenwar: Please do ignore it. I misread the "||" operator. 
It should take care of my concern.

Issue #179, opened 2019-11-20
[v1alpha2] Make setError functions generic


Xenwar:
The value of Status.ErrorMessage and Status.ErrorReason need to be set in the controller code. To this end, four SetError* methods are created, one for each Object type. We need to centralize the error related code as it is easier to maintain. 

**To Do:**

- Create a folder for error related parts (much like` cluster-api/errors/`)
- Define new error constants specific to BareMetalMachine and clusters
- Define generic SetError and ClearError functions that all four objects in a generic way.

Xenwar:
This issue is no longer relevant. The code has changed significantly for this approach to be taken into use.

PR #191, opened 2019-11-28
Changed error setting function signature and reduced verbosity of err…

Xenwar:
This commit make error setting functions to take the same parameters across controllers. 
It also reduce verbosity of error messages that are set on statuses.

### DIFF
--- a/baremetal/baremetalcluster_manager.go
+++ b/baremetal/baremetalcluster_manager.go
@@ -108,7 +108,7 @@ func (s *ClusterManager) Create(ctx context.Context) error {
 	err := config.IsValid()
 	if err != nil {
 		// Should have been picked earlier. Do not requeue
-		s.setError(err.Error(), capierrors.InvalidConfigurationClusterError)
+		s.setError("Invalid BareMetalCluster provided", capierrors.InvalidConfigurationClusterError)
 		return err
 	}
 
@@ -162,7 +162,7 @@ func (s *ClusterManager) UpdateClusterStatus() error {
 	// Get APIEndpoints from  BaremetalCluster Spec
 	endpoints, err := s.apiEndpoints()
 	if err != nil {
-		s.setError(err.Error(), capierrors.InvalidConfigurationClusterError)
+		s.setError("Invalid APIEndpoints values", capierrors.InvalidConfigurationClusterError)
 		return err
 	}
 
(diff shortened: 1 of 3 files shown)


baremetal/baremetalcluster_manager.go
```diff
@@ -162,7 +163,8 @@ func (s *ClusterManager) UpdateClusterStatus() error {
 	// Get APIEndpoints from  BaremetalCluster Spec
 	endpoints, err := s.apiEndpoints()
 	if err != nil {
-		s.setError(err.Error(), capierrors.InvalidConfigurationClusterError)
+		msg := "Invalid APIEndpoints values"
```
kashifest: There is now an APIEndPoint error type defined in baremetalcluster_types. Should we use that ?

https://github.com/metal3-io/cluster-api-provider-baremetal/blob/f0b9abda683aec54034177c3ad21bbc3fa6a6dcf/api/v1alpha2/baremetalcluster_types.go#L41

Xenwar: Currently, the error setting functions take only ClusterStatusError or MachineStatusError as parameter. There is an issue for changing the function signatures so that they take a generic higher level type. https://github.com/metal3-io/cluster-api-provider-baremetal/issues/179

baremetal/baremetalmachine_manager.go
```diff
@@ -202,7 +202,8 @@ func (m *MachineManager) Associate(ctx context.Context) error {
 	err := config.IsValid()
 	if err != nil {
 		// Should have been picked earlier. Do not requeue
-		m.setError(ctx, err.Error(), capierrors.InvalidConfigurationMachineError)
+		msg := "BareMetalMachine has invalid configuration"
+		m.setError(ctx, msg, capierrors.InvalidConfigurationMachineError)
```
maelk: done in #187

baremetal/baremetalcluster_manager.go
```diff
@@ -108,7 +108,8 @@ func (s *ClusterManager) Create(ctx context.Context) error {
 	err := config.IsValid()
 	if err != nil {
 		// Should have been picked earlier. Do not requeue
-		s.setError(err.Error(), capierrors.InvalidConfigurationClusterError)
+		msg := "Invalid BareMetalCluster provided"
+		s.setError(msg, capierrors.InvalidConfigurationClusterError)
```
maelk: you can define the msg inline

Issue #189, opened 2019-11-26
[v1alpha2] flake on TestAssociate for baremetal machine manager


maelk:
The TestAssociate case is flaky : 

```
 --- FAIL: TestAssociate (0.00s)
    baremetalmachine_manager_test.go:1620: ## TC-Associate machine, host empty, baremetal machine spec set ##
    baremetalmachine_manager_test.go:1636: requeue in: 30s
    baremetalmachine_manager_test.go:1620: ## TC-Associate empty machine, baremetal machine spec nil ##
    baremetalmachine_manager_test.go:1620: ## TC-Associate empty machine, baremetal machine spec set ##
    baremetalmachine_manager_test.go:1620: ## TC-Associate empty machine, host empty, baremetal machine spec set ## 
```

Jaakko-Os:
I'll look into this

maelk:
fixed

PR #192, opened 2019-11-29
fix BMM flaky TestAssociate unit test

Jaakko-Os:
Closes: #189

### DIFF
--- a/baremetal/baremetalmachine_manager_test.go
+++ b/baremetal/baremetalmachine_manager_test.go
@@ -27,6 +27,7 @@ import (
 
 	bmoapis "github.com/metal3-io/baremetal-operator/pkg/apis"
 	bmh "github.com/metal3-io/baremetal-operator/pkg/apis/metal3/v1alpha1"
+	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


baremetal/baremetalmachine_manager_test.go
```diff
@@ -1419,7 +1424,9 @@ func TestAssociate(t *testing.T) {
 
 		err = machineMgr.Associate(context.TODO())
 		if err != nil {
-			t.Error(err)
+			if !tc.ExpectRequeue {
+				t.Error(err)
+			}
```
maelk: this is too simplistic. Having an error does not always mean that we have a requeue. You should separate expectError and expectRequeue. And if you expect an error and don't get it ? you should also have a condition if err is nil and you expect error, then t.Error(...). If you do the proper testing on this condition, I expect that the flake will not be fixed, since the test was sometimes passing (i.e. no requeue) and sometimes not passing (i.e. requeue). If you set expectRequeue, then you'll just flip the side, but if the test is flaky, then you'll get errors when you are not getting a requeue but were expecting one. This PR now addresses only the symptoms, and rather hides them than solving the actual issue with the test.

baremetal/baremetalmachine_manager_test.go
```diff
@@ -1376,29 +1376,34 @@ func TestSetNodeProviderID(t *testing.T) {
...
+			Machine:       newMachine("", "", nil),
+			BMMachine:     newBareMetalMachine("mybmmachine", nil, bmmSpecAll, nil, bmmObjectMetaWithValidAnnotations),
+			Host:          newBareMetalHost("", nil, bmh.StateNone, nil, false),
+			ExpectRequeue: false,
 		},
 		"Associate machine, host empty, baremetal machine spec set": {
-			Machine:   newMachine("mymachine", "", nil),
-			BMMachine: newBareMetalMachine("mybmmachine", nil, bmmSpecAll, nil, bmmObjectMetaWithValidAnnotations),
-			Host:      newBareMetalHost("", nil, bmh.StateNone, nil, false),
+			Machine:       newMachine("mymachine", "", nil),
```
Xenwar: I got the impression the name was the issue.
Would adding ExpectedRequeue value prevent the issue ?
If it does not stop it, then using unique names for resources might help.

baremetal/baremetalmachine_manager_test.go
Jaakko-Os: I was troubleshooting chooseHost() which is executed during the flaky test. That function returns host and err. If host is nil -> requeue happens. 
Anyways the testcase was run 20+ times continuously and there were no failures. Of course the test case is good to be strict 
I think https://github.com/metal3-io/cluster-api-provider-baremetal/pull/187 implements so many changes that this issue might be obsolete.

maelk: Unfortunately, #187 also hits the issue

Jaakko-Os: The failing test was depending on the possible earlier tests run within TestAssociate.
The host was not set correctly to nil. Fix sets the host to nil.

baremetal/baremetalmachine_manager_test.go
Jaakko-Os: ExpectedRequeue used

baremetal/baremetalmachine_manager_test.go
```diff
@@ -1376,50 +1376,66 @@ func TestSetNodeProviderID(t *testing.T) {
...
-		err = machineMgr.Associate(context.TODO())
-		if err != nil {
-			t.Error(err)
+		// retVal can contain an error or a requeue value
+		retVal := machineMgr.Associate(context.TODO())
+		if retVal != nil {
+			if !tc.ExpectRequeue {
+				t.Error(err)
+			}
 		}
```
maelk: this is not the correct way to test it. you should check the type of the error, if it is a requeue, then check if you were expecting it. if not error. If you were expecting a requeue and did not get one, you should also send an error. Then if the error is not a requeue, you were probably not expecting an error, then send an error

baremetal/baremetalmachine_manager_test.go
```diff
@@ -1376,50 +1376,66 @@ func TestSetNodeProviderID(t *testing.T) {
 
 func TestAssociate(t *testing.T) {
 	testCases := map[string]struct {
-		Machine   *capi.Machine
-		Host      *bmh.BareMetalHost
-		BMMachine *capbm.BareMetalMachine
+		Machine       *capi.Machine
+		Host          *bmh.BareMetalHost
+		BMMachine     *capbm.BareMetalMachine
+		Objects       []runtime.Object
```
maelk: no need to make it an object of the test case

baremetal/baremetalmachine_manager_test.go
```diff
@@ -1376,50 +1376,66 @@ func TestSetNodeProviderID(t *testing.T) {
...
 
 	for name, tc := range testCases {
 		t.Logf("## TC-%s ##", name)
 
-		objects := []runtime.Object{
-			tc.Host,
-			tc.BMMachine,
-			tc.Machine,
+		if tc.Host == nil {
+			tc.Objects = []runtime.Object{
```
maelk: you can still make it an object that is not from the test case

baremetal/baremetalmachine_manager_test.go
Jaakko-Os: fixed

baremetal/baremetalmachine_manager_test.go
Jaakko-Os: done

baremetal/baremetalmachine_manager_test.go
Jaakko-Os: no change made here

baremetal/baremetalmachine_manager_test.go
```diff
@@ -1633,7 +1647,14 @@ func TestAssociate(t *testing.T) {
...
 		if err != nil {
-			t.Error(err)
+			if !tc.ExpectRequeue {
+				t.Error(err)
+			} else {
+				requeueErr, ok := errors.Cause(err).(HasRequeueAfterError)
+				if !ok {
+					t.Error(requeueErr)
+				}
+			}
```
maelk: what if err is nil but you were expecting a requeue ?

Jaakko-Os: good point. condition added.

PR #195, opened 2019-12-02
fixes stuck deletion when host has no Status

mhrivnak:
If a Machine is being deleted and is associated with a BareMetalHost that has
no status, the deletion logic would fail to identify that the host is
unprovisioned. It would requeue forever, waiting for the host to reach an
unprovisioned state.

This can happen if there are BareMetalHosts without a running
baremetal-operator, and Machines with this provider running.


maelk:
I have made the v1alpha2 equivalent in #196

PR #198, opened 2019-12-03
Refactor the error tests for the managers

maelk:
use ginkgo and gomega to uniformize the tests

### DIFF
--- a/baremetal/requeue_error_test.go
+++ b/baremetal/requeue_error_test.go
@@ -18,8 +18,10 @@ package baremetal
 
 import (
 	"fmt"
-	"testing"
 	"time"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


baremetal/requeue_error_test.go
```diff
@@ -18,28 +18,26 @@ package baremetal
...
-	if err.GetRequeueAfter() != duration {
-		t.Errorf("Error in duration, expected %vs, got %s", RequeueDuration2,
-			err.RequeueAfter,
-		)
-	}
-}
+var _ = Describe("Errors testing", func() {
+	It("returns the correct error", func() {
+		err := &RequeueAfterError{time.Second * RequeueDuration1}
+		Expect(err.Error()).To(Equal(fmt.Sprintf("requeue in: %vs", RequeueDuration1)))
```
kashifest: Checking the error string isn't particularly a good practice. Can we compare it with an error type?

maelk: we are testing err.Error(), that returns the error string so we need to compare the error string

PR #204, opened 2019-12-11
Minor fixes for the tests

maelk:
- improve the objects creation to avoid test pollution
- use assertion NotTo(HaveOccurred()) for errors instead of To(Succeed())


kashifest:
looks good to me

PR #206, opened 2019-12-11
Fix image version for deployment

maelk:
switch to v1alpha1 for capbm image tag


maelk:
The unit tests will fail until the trigger to build capbm-unit:v1alpha1 will be set. working on it.

PR #211, opened 2019-12-17
adds auto-scaling for MachineSet based on number of BareMetalHosts

### LINKED ISSUE #188: proposal: add auto-scaling for MachineSets
In many clusters, it is desirable for the size of a MachineSet to always equal the number of matching BareMetalHosts. In such a scenario, the cluster owner wants all of their hardware to be provisioned and turned into Nodes, and they want to remove excess Machines in case they remove hosts from their cluster.

Rather than make some external process manage the size of MachineSets as BareMetalHosts come and go, we could create a small controller that (optionally) automatically ensures a MachineSet has a size equal to the number of matching BareMetalHosts.

The controller would be an additional [`Controller`](https://github.com/metal3-io/cluster-api-provider-baremetal/blob/134c3b78ec636f79fd6703ea273dbe0877730ad4/vendor/sigs.k8s.io/controller-runtime/pkg/controller/controller.go#L44:6) in this project. It would watch MachineSets as its primary resource, and if they have a particular annotation, ensure that their size equals the number of matching BareMetalHosts. It would watch BareMetalHosts as a secondary resource.

Thoughts?

mhrivnak:
In many clusters, it is desirable for the size of a MachineSet to always equal
the number of matching BareMetalHosts. In such a scenario, the cluster owner
wants all of their hardware to be provisioned and turned into Nodes, and they
want to remove excess Machines in case they remove hosts from their cluster.
This change adds a controller that scales MachineSets to the number of matching
BareMetalHosts. The behavior is opt-in, requiring an annotation on the
MachineSet.

fixes #188

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,3 +20,9 @@ For more information about this actuator and related repositories, see
 
 See the [API Documentation](docs/api.md) for details about the `providerSpec`
 API used with this `cluster-api` provider.
+
+## MachineSet Scaling
+
+If you would like a MachineSet to be automatically scaled to the number of
+matching BareMetalHosts, annotate that MachineSet with key
+`metal3.io/autoscale-to-hosts` and any value.
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -25,6 +25,7 @@ import (
 	bmoapis "github.com/metal3-io/baremetal-operator/pkg/apis"
 	"github.com/metal3-io/cluster-api-provider-baremetal/pkg/apis"
 	"github.com/metal3-io/cluster-api-provider-baremetal/pkg/cloud/baremetal/actuators/machine"
+	"github.com/metal3-io/cluster-api-provider-baremetal/pkg/controller"
 	"github.com/metal3-io/cluster-api-provider-baremetal/pkg/manager/wrapper"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/discovery"
@@ -96,6 +97,11 @@ func main() {
 
 	// the manager wrapper will add an extra Watch to the controller
 	capimachine.AddWithActuator(wrapper.New(mgr), machineActuator)
+
+	if err := controller.AddToManager(mgr); err != nil {
+		log.Error(err, "Failed to add controller to manager")
+		os.Exit(1)
+	}
 
 	if err := mgr.Start(signals.SetupSignalHandler()); err != nil {
 		entryLog.Error(err, "unable to run manager")
--- a/pkg/controller/machineset/machineset_controller.go
+++ b/pkg/controller/machineset/machineset_controller.go
@@ -1,0 +1,156 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package machineset
+
+import (
+	"context"
+
+	bmh "github.com/metal3-io/baremetal-operator/pkg/apis/metal3/v1alpha1"
+	actuator "github.com/metal3-io/cluster-api-provider-baremetal/pkg/cloud/baremetal/actuators/machine"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
+	machinev1alpha1 "sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+var log = logf.Log.WithName("machineset-controller")
+
+// Add creates a new MachineSet Controller and adds it to the Manager with default RBAC. The Manager will set fields on the Controller
+// and Start it when the Manager is Started.
+func Add(mgr manager.Manager) error {
+	return add(mgr, newReconciler(mgr))
+}
+
+// newReconciler returns a new reconcile.Reconciler
+func newReconciler(mgr manager.Manager) reconcile.Reconciler {
+	return &ReconcileMachineSet{Client: mgr.GetClient(), scheme: mgr.GetScheme()}
+}
+
+// add adds a new Controller to mgr with r as the reconcile.Reconciler
+func add(mgr manager.Manager, r reconcile.Reconciler) error {
+	// Create a new controller
+	c, err := controller.New("machineset-controller", mgr, controller.Options{Reconciler: r})
+	if err != nil {
+		return err
+	}
+
+	// Watch for changes to MachineSet
+	err = c.Watch(&source.Kind{Type: &machinev1alpha1.MachineSet{}},
+		&handler.EnqueueRequestForObject{}, predicate.ResourceVersionChangedPredicate{})
+	if err != nil {
+		return err
+	}
+
+	mapper := msmapper{client: mgr.GetClient()}
+	err = c.Watch(&source.Kind{Type: &bmh.BareMetalHost{}},
+		&handler.EnqueueRequestsFromMapFunc{ToRequests: &mapper}, predicate.ResourceVersionChangedPredicate{})
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+var _ reconcile.Reconciler = &ReconcileMachineSet{}
+
+// ReconcileMachineSet reconciles a MachineSet object
+type ReconcileMachineSet struct {
+	client.Client
+	scheme *runtime.Scheme
+}
+
+// Reconcile reads that state of the cluster for a MachineSet object and makes changes based on the state read
+// and what is in the MachineSet.Spec
+// Automatically generate RBAC rules to allow the Controller to read and write Deployments
+// +kubebuilder:rbac:groups=cluster.k8s.io,resources=machinesets,verbs=get;list;watch;update;patch
+func (r *ReconcileMachineSet) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	log := log.WithValues("MachineSet", request.NamespacedName.String())
+	ctx := context.TODO()
+	// Fetch the MachineSet instance
+	instance := &machinev1alpha1.MachineSet{}
+	err := r.Get(context.TODO(), request.NamespacedName, instance)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			// Object not found, return.  Created objects are automatically garbage collected.
+			// For additional cleanup logic use finalizers.
+			return reconcile.Result{}, nil
+		}
+		// Error reading the object - requeue the request.
+		return reconcile.Result{}, err
+	}
+
+	// Don't take action if the annotation is not present
+	annotations := instance.ObjectMeta.GetAnnotations()
+	if annotations == nil {
+		return reconcile.Result{}, nil
+	}
+	_, present := annotations[AutoScaleAnnotation]
+	if !present {
+		return reconcile.Result{}, nil
+	}
+
+	selector, err := actuator.SelectorFromProviderSpec(&instance.Spec.Template.Spec.ProviderSpec)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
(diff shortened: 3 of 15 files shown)


README.md
```diff
@@ -20,3 +20,9 @@ For more information about this actuator and related repositories, see
 
 See the [API Documentation](docs/api.md) for details about the `providerSpec`
 API used with this `cluster-api` provider.
+
+## MachineSet Scaling
+
+If you would like a MachineSet to be automatically scaled to the number of
+matching BareMetalHosts, annotate that MachineSet with key
+`metal3.io/autoscaletohosts` and any value.
```
dhellmann: readability nit: Is it usual to use annotation names with multiple words and no separating punctuation?

pkg/controller/machineset/machineset_controller.go
```diff
@@ -0,0 +1,136 @@
...
+		}
+		// Error reading the object - requeue the request.
+		return reconcile.Result{}, err
+	}
+
+	selector, err := actuator.SelectorFromProviderSpec(&instance.Spec.Template.Spec.ProviderSpec)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
```
dhellmann: I see the logic that looks for the annotation in mapper, but given the watch set up on line 60 I think we need a check here, too, don't we?

pkg/controller/machineset/machineset_controller_test.go
```diff
@@ -0,0 +1,150 @@
...
+		HostSelector: bmv1alpha1.HostSelector{
+			MatchLabels: map[string]string{"size": "large"},
+		},
+	})
+	g.Expect(err).NotTo(gomega.HaveOccurred())
+	instance := &machinev1alpha1.MachineSet{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        "machineset1",
+			Namespace:   "default",
+			Annotations: map[string]string{AutoScaleAnnotation: "yesplease"},
```
dhellmann: It would be good to have a test to show that a MachineSet without the annotation is ignored.

README.md
mhrivnak: I'm not sure there's much of a convention.

The example here is `imageregistry`: https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set

But I also see annotation and label keys hyphenated sometimes. We could make this one `metal3.io/autoscale-to-hosts`. WDYT? I don't have a preference.

pkg/controller/machineset/machineset_controller.go
mhrivnak: Indeed! Will add that.

README.md
dhellmann: I like the hyphenated version better. Although given my track record, that probably means it's not the standard. ;-)

mhrivnak:
@dhellmann this is ready for re-review.

pkg/controller/machineset/machineset_controller.go
```diff
@@ -0,0 +1,146 @@
...
+	var count int32
+	for _, host := range hosts.Items {
+		if selector.Matches(labels.Set(host.ObjectMeta.Labels)) {
+			count++
+		}
+	}
+
+	if instance.Spec.Replicas == nil || count != *instance.Spec.Replicas {
+		log.Info("Scaling MachineSet", "replicas", count)
```
dhellmann: It might be nice to include the old count value here in the debug output.

pkg/controller/machineset/machineset_controller.go
```diff
@@ -0,0 +1,146 @@
...
+		new := instance.DeepCopy()
+		new.Spec.Replicas = &count
+		err = r.Update(ctx, new)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
```
dhellmann: This is going to cause machine sets to scale down, too. I think we want that, but I also think we want to try to select the right machines. For example, if I define a MachineSet with a set of label selectors and then later change those selectors, the set of hosts that match are likely to change. We should ensure that only hosts that do not match the labels are retained and any hosts that do not match are deprovisioned. I think we can do that by adding an annotation to the machine to select that machine for deletion. We can do this in a separate PR, if you prefer.

pkg/controller/machineset/mapper.go
```diff
@@ -0,0 +1,70 @@
...
+			if !present {
+				continue
+			}
+
+			matches, err := m.hostMatchesMachineSet(host, &ms)
+			if err != nil {
+				return []reconcile.Request{}
+			}
+			if matches {
+				return []reconcile.Request{
```
dhellmann: What if a host matches more than one machine set?

cmd/manager/main.go
```diff
@@ -97,6 +98,11 @@ func main() {
 	// the manager wrapper will add an extra Watch to the controller
 	capimachine.AddWithActuator(wrapper.New(mgr), machineActuator)
 
+	if err := controller.AddToManager(mgr); err != nil {
+		log.Error(err, "")
```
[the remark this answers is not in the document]
mhrivnak: Good point. I think that's what operator-sdk scaffolded out, and I probably just didn't look at it. I'm changing the string to "Failed to add controller to manager".

pkg/controller/machineset/machineset_controller.go
mhrivnak: Quick disambiguation for anyone that finds it helpful: The label selector that could be eligible for change is in the ProviderSpec (which is embedded in the MachineSet's spec), and it expresses how Machines match BareMetalHosts. That is different from the label selector in the MachineSet's own spec, which is how it keeps track of its Machines and should not be changed.

That said, I think the selectors in the ProviderSpec should be considered immutable. Any Machines that already exist for a MachineSet are not going to have their ProviderSpec updated if it changes (the [MachineSet controller](https://github.com/kubernetes-sigs/cluster-api/blob/release-0.1/pkg/controller/machineset/machineset_controller.go) just doesn't do that). That would result in the MachineSet having an exciting and unpredictable mix of Machines that match different groups of BareMetalHosts. If you need to change the ProviderSpec's HostSelector, it's better to make a new MachineSet. It might be wise to make a validating admission webhook that enforces this.

For deletion, the safe steps are documented [here](https://github.com/metal3-io/metal3-docs/blob/master/design/remove-host.md). This PR automates the "Scale down the MachineSet" portion, and I agree there is an opportunity to also locally automate the addition of the `cluster.k8s.io/delete-machine` annotation to a Machine whose BareMetalHost is being deleted or no longer matches. Probably the right approach would be for the Machine controller (actuator) to verify on each reconcile that the BareMetalHost still matches and isn't being deleted; otherwise it breaks the association . +1 to doing that in a separate PR.

pkg/controller/machineset/mapper.go
mhrivnak: This feature is only useful in a situation where the MachineSet has exclusive access to the BareMetalHosts its Machines will match. If two MachineSets shared a pool of BareMetalHosts, and both of them used this scaling feature, there would be twice as many Machines created as there are BareMetalHosts, and it would put the two MachineSets in a race to grab as many of the hosts as they could before they run out.

That said, I could scan the whole list here instead of returning at the first match, and log a warning in case there are multiple matches. Though I suspect the end user would figure out their misstep regardless. WDYT?

mhrivnak:
I'm currently chasing a nil pointer dereference that *seems* unrelated to this PR, but we all know how that can turn out, so I want to get to the bottom of it before merging.

mhrivnak:
I fixed the nil pointer dereference, so this is safe to merge. It ended up being a result of [this landmine](https://github.com/metal3-io/cluster-api-provider-baremetal/blob/1a6848c7e5d7ab9c7357e7468862520884e8a132/pkg/controller/add_machine.go#L29) leftover from the earliest days of scaffolding the actuator. It wasn't on any code path, until now. The "real" actuator gets created only [in main.go](https://github.com/metal3-io/cluster-api-provider-baremetal/blob/1a6848c7e5d7ab9c7357e7468862520884e8a132/cmd/manager/main.go#L58-L60). Because of how both kubebuilder and operator-sdk scaffold things, the abandoned code was reactivated when I added the machineset controller in the same package.

pkg/controller/machineset/machineset_controller_test.go
mhrivnak: Done.

cmd/manager/main.go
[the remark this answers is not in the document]
mhrivnak: Done.

pkg/controller/machineset/machineset_controller.go
mhrivnak: Done.

pkg/controller/machineset/machineset_controller.go
mhrivnak: You make a good point, though I think you're identifying a problem that goes beyond this feature. Even without considering this PR, it sounds like a bad idea for a production cluster to have a MachineSet whose ProviderSpec has an empty HostSelector. That makes it logistically challenging to ever add a second MachineSet that targets specific hosts, because you'll always have that original MachineSet with its Machines that could match anything. Maybe we should ensure that the default MachineSet gets created with a default HostSelector, and that BareMetalHosts get a corresponding label by default.

dhellmann: @stbenjam @zaneb @russellb I think we could use some additional input on this.

pkg/controller/machineset/mapper.go
mhrivnak: With OpenShift, apparently if a Machine goes too long without being turned into a Node, it results in a [critical-severity alert](https://github.com/openshift/machine-api-operator/blob/57f529071966836be0cbb1bebc531f191e843691/install/0000_90_machine-api-operator_04_alertrules.yaml). That'll happen whether someone manually scales a MachineSet too large, or incorrectly uses automation to scale their MachineSets.

I'm thinking the best behavior here is to reconcile however many MachineSets would match a host, and during reconciliation, log a warning if a MachineSet has the annotation but does not have a HostSelector. When the alarms go off, if they look at the log output, that should help them quickly figure out the problem.

dhellmann: I sort of like the idea, because it relies on an existing alert. Do we expect users to need to go to the logs to figure out situations like this, though? We can predict this as a source of user error; can we devise a way to make the cause of the error obvious? What if the autoscaler annotated a host with all of the machine sets it matched, for example?

mhrivnak: Some alerts explicitly say in their text to look in the logs for a particular pod, but not this one. It would be interesting to know what @andybraren thinks about how we can best lead the user to a solution.

The problem will be one of two things:
1. a Machine was matched with a host, but that host is failing to provision. Hopefully the host itself is able to represent useful information about the failure, and that should make the source of the problem apparent. Perhaps the host's failure generates an additional alert?
2. a MachineSet is over-sized. Either add some hosts or reduce the MachineSet's size. In most cases we can avoid this problem by automatically adjusting the MachineSet size, but that opens the opportunity for someone to have two HostSelectors that overlap, per the above conversation. For that (I suspect narrow) case, they really should fix the selectors so that there are not overlapping MachineSets. @andybraren any thoughts on how best to surface that problem to the user?

mhrivnak: > I'm thinking the best behavior here is to reconcile however many MachineSets would match a host, and during reconciliation, log a warning if a MachineSet has the annotation but does not have a HostSelector.

I made these changes.

andybraren: Would it be possible & appropriate to generate an error event when a Machine is unable to bind to a host for this reason? That feedback would appear almost immediately in the Web Console, and help users connect the dots between the time they set the annotation and the `MachineWithoutValidNode` alert firing ~10 minutes later.

If the event's `involvedObject.kind` is the same machine that triggers the alert, it would be quite easy to trace the issue in the Web Console UI at least.

And if we can't create an event, under normal circumstances it's pretty unlikely that a user could opt into this behavior and somehow _not_ trigger the `MachineWithoutValidNode` alert ~10 minutes later if there's a collision, correct? If the alert could trigger hours or days later I'd be more concerned, but 10 minutes should be close enough to correlate (although some SREs might get annoyed).

andybraren: > a Machine was matched with a host, but that host is failing to provision. Hopefully the host itself is able to represent useful information about the failure, and that should make the source of the problem apparent. Perhaps the host's failure generates an additional alert?

Agreed, it sounds like enough context would be provided by the host's status in most cases, but it'd probably be nice if a failed provision generated an error event as well for easier problem tracing.

mhrivnak: It is definitely possible to create warning events for any of these situations. Whether it's appropriate is a much harder question!

If a Machine doesn't get matched with a host, it will retry every 30s. We probably don't want to create the same event every 30s, unless that's a normal pattern with the console? There are also several scenarios where a host might not be available right now, but in a few minutes it will be. Is it warning-worthy to tell the user that a suitable host could not be found? For that reason, I think the 10-minute delay is useful. Maybe it would help for the actuator to create a "Normal" event that describes that it could not be matched with a host; that way the event won't set off any alarms, but if someone looks at the Machine wondering why it hasn't turned into a Node, they'll see the events explaining why.

If a user has a MachineSet with no host selector, we may want to warn them about that. But again, do we want to warn them every time we reconcile that MachineSet? Probably not.

Creating an event is a better fit for when something happened, and in the case of a controller, we usually only know that something happened if we're the ones who did it (or tried and failed to do it). Controllers deliberately avoid basing logic around actions, and instead focus exclusively on comparing current state vs. desired state. They don't care how the current state came to be; they just want to figure out what changes to make to bring the current state closer to the desired state.

IMO trying to match a machine to a host qualifies as the controller doing something and is a candidate for generating an event. Noticing that a MachineSet is being misused is not such a neat scenario.

mhrivnak:
Moving this PR to the openshift repo: https://github.com/openshift/cluster-api-provider-baremetal/pull/52

PR #225, opened 2020-01-21
Naive merge of machine remediation code

beekhof:
There is scope to optimise the directory layout, but this should be enough to comment on


pkg/baremetal/remediator/remediator.go
```diff
@@ -0,0 +1,322 @@
...
+
+	now := time.Now()
+	switch machineRemediation.Status.State {
+	// initiating the reboot action
+	case mrv1.RemediationStateStarted:
+		rebootInProgress := isRebootInProgress(bmh)
+		// skip the reboot in case when the machine has power off state before the reboot action
+		// it can mean that an user power off the machine by purpose
+		if !bmh.Spec.Online && !rebootInProgress {
+			glog.V(4).Infof("Skip the remediation, machine %q has power off state before the remediation action", machine.Name)
```
dhellmann: This smells like a race condition. What would power the host off before remediation starts?

pkg/baremetal/remediator/remediator.go
```diff
@@ -0,0 +1,322 @@
...
+			bmr.recorder.Eventf(
+				machine,
+				corev1.EventTypeNormal,
+				"MachineRemediationSkippedOffline",
+				"Remediation of machine %q skipped because it was in power off state already",
+				machine.Name,
+			)
+			mrCopy.Status.State = mrv1.RemediationStateSucceeded
+			mrCopy.Status.Reason = "Skip the reboot, the machine power off by an user"
```
dhellmann:
```suggestion
			mrCopy.Status.Reason = "Skip the reboot, the machine was powered off already"
```

pkg/baremetal/remediator/remediator.go
```diff
@@ -0,0 +1,322 @@
...
+				corev1.EventTypeNormal,
+				"MachineRemediationRebootStarted",
+				"Reboot of machine %q has started",
+				machine.Name,
+			)
+
+			mrCopy.Status.State = mrv1.RemediationStatePowerOff
+			mrCopy.Status.Reason = "Starts the reboot process"
+		}
+		return bmr.client.Status().Update(context.TODO(), mrCopy)
```
dhellmann: What happens if this call fails? The states between the 2 resources (host and remediation) will be out of sync. Is there enough logic here to handle that case?

Could you you could extend the state stored on the host so it could serve as the source of truth, and then just reflect it in the remediation resource status fields? Would that make the controller simpler? Would that also help with races between multiple remediation requests and avoid the need to check the power state and rebootInProgress?

pkg/apis/machineremediation/register.go
```diff
@@ -0,0 +1,4 @@
+package machineremediation
+
+// GroupName contains the name of the API group
+const GroupName = "machineremediation.kubevirt.io"
```
dhellmann: I think we want this to be metal3 not KubeVirt.

```suggestion
const GroupName = "machineremediation.metal3.io"
```

pkg/apis/machineremediation/v1alpha1/doc.go
```diff
@@ -0,0 +1,4 @@
+// Package v1alpha1 contains API Schema definitions for the healthchecking v1alpha1 API group
+// +k8s:deepcopy-gen=package,register
+// +groupName=machineremediation.kubevirt.io
```
dhellmann:
```suggestion
// +groupName=machineremediation.metal3.io
```

pkg/apis/machineremediation/v1alpha1/machineremediation_types.go
```diff
@@ -0,0 +1,74 @@
...
+// MachineRemediationSpec defines the spec of MachineRemediation
+type MachineRemediationSpec struct {
+	// Type contains the type of the remediation
+	Type RemediationType `json:"type,omitempty" valid:"required"`
+	// MachineName contains the name of machine that should be remediate
+	MachineName string `json:"machineName,omitempty" valid:"required"`
+}
+
+// MachineRemediationStatus defines the observed status of MachineRemediation
+type MachineRemediationStatus struct {
```
dhellmann: I think you're going to want to use annotations on the host to serve as the source of truth for at least some of these status values, to avoid race conditions between updating the host state and then the MR state. At least State and possibly the time values.

pkg/baremetal/remediator/remediator.go
```diff
@@ -0,0 +1,322 @@
...
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/record"
+
+	mapiv1 "sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+)
+
+const (
+	rebootDefaultTimeout = 5
```
dhellmann: This value is converted to minutes in 2 places in the code below. It could be converted once here so that the reader can tell that it means 5 minutes.

pkg/apis/machineremediation/v1alpha1/machineremediation_types.go
```diff
@@ -0,0 +1,74 @@
...
+	State     RemediationState `json:"state,omitempty"`
+	Reason    string           `json:"reason,omitempty"`
```
dhellmann: I would expect Reason to be part of the spec and something we let the user set. It looks like this field is used for a human readable status string, not really the reason for the remediation?

pkg/baremetal/remediator/remediator.go
```diff
@@ -0,0 +1,322 @@
...
+			)
+
+			mrCopy.Status.State = mrv1.RemediationStatePowerOff
+			mrCopy.Status.Reason = "Starts the reboot process"
+		}
+		return bmr.client.Status().Update(context.TODO(), mrCopy)
+
+	case mrv1.RemediationStatePowerOff:
+		// failed the remediation on timeout
+		if machineRemediation.Status.StartTime.Time.Add(rebootDefaultTimeout * time.Minute).Before(now) {
```
dhellmann: If we're always powering off the host, how would we end up in this state? Failure of the baremetal controller to power the host back on?

pkg/baremetal/remediator/remediator.go
```diff
@@ -0,0 +1,322 @@
...
+			machine.Name,
+		)
+
+		mrCopy.Status.State = mrv1.RemediationStatePowerOn
+		mrCopy.Status.Reason = "Reboot in progress"
+		return bmr.client.Status().Update(context.TODO(), mrCopy)
+
+	case mrv1.RemediationStatePowerOn:
+		// failed the remediation on timeout
+		if machineRemediation.Status.StartTime.Time.Add(rebootDefaultTimeout * time.Minute).Before(now) {
```
dhellmann: This logic looks very similar to the logic that starts the power off state handling block. Could this come before the switch statement so the code only appears once?

pkg/baremetal/remediator/remediator.go
```diff
@@ -0,0 +1,322 @@
...
+			mrCopy.Status.EndTime = &metav1.Time{Time: now}
+			return bmr.client.Status().Update(context.TODO(), mrCopy)
+		}
+
+		node, err := getNodeByMachine(bmr.client, machine)
+		if err != nil {
+			// we want to reconcile with delay of 10 seconds when the machine does not have node reference
+			// or node does not exist
+			if errors.IsNotFound(err) {
+				glog.Warningf("The machine %q node does not exist", machine.Name)
```
dhellmann: I'm not sure this should be a warning, if we expect it to happen. Is there anything in the remediation CR reflecting the fact that we are waiting for a node to be recreated?

pkg/baremetal/remediator/remediator.go
```diff
@@ -0,0 +1,322 @@
...
+			mrCopy.Status.Reason = "Reboot succeeded"
+			mrCopy.Status.EndTime = &metav1.Time{Time: now}
+			return bmr.client.Status().Update(context.TODO(), mrCopy)
+		}
+		return nil
+
+	// assumption that the reboot annotation removed because of node removal
+	case mrv1.RemediationStateSucceeded:
+		// remove machine remediation object
+		return bmr.client.Delete(context.TODO(), machineRemediation)
```
dhellmann: Is this just a garbage collection mechanism? Is it usual to have a controller delete its instructions like this?

pkg/components/deployments.go
```diff
@@ -0,0 +1,143 @@
+package components
```
dhellmann: What is this file for?

pkg/components/rbac.go
```diff
@@ -0,0 +1,162 @@
...
+	mrv1 "github.com/metal3-io/cluster-api-provider-baremetal/pkg/apis/machineremediation/v1alpha1"
+)
+
+var (
+	// Rules contains rules for all machine remediation components
+	Rules = map[string][]rbacv1.PolicyRule{
+		ComponentMachineRemediation: []rbacv1.PolicyRule{
+			{
+				APIGroups: []string{
+					"machine.openshift.io",
```
dhellmann: This looks openshift-specific?

pkg/components/rbac.go
```diff
@@ -0,0 +1,162 @@
...
+				},
+				Verbs: []string{
+					"get",
+					"list",
+					"watch",
+				},
+			},
+			{
+				APIGroups: []string{
+					"machineremediation.kubevirt.io",
```
dhellmann:
```suggestion
					"machineremediation.metal3.io",
```

pkg/consts/consts.go
```diff
@@ -0,0 +1,23 @@
+package consts
+
+const (
+	// AnnotationBareMetalHost contains the annotation key for bare metal host
+	AnnotationBareMetalHost = "metal3.io/BareMetalHost"
+	// AnnotationMachine contains the annotation key for machine
+	AnnotationMachine = "machine.openshift.io/machine"
```
dhellmann: I think OpenShift uses a different name here?

pkg/consts/consts.go
```diff
@@ -0,0 +1,23 @@
...
+	// AnnotationNodeMachineReboot contains machine reboot annotation key, once nodereboot controller will detect it,
+	// it will create the MachineRemediation object
+	AnnotationNodeMachineReboot = "healthchecking.openshift.io/machine-remediation-reboot"
```
dhellmann: OpenShift

pkg/consts/consts.go
```diff
@@ -0,0 +1,23 @@
...
+	// AnnotationRebootInProgress contains the annotation key, that indicates that reboot in the progress
+	AnnotationRebootInProgress = "machineremediation.kubevirt.io/rebootInProgress"
```
dhellmann:
```suggestion
	AnnotationRebootInProgress = "machineremediation.metal3.io/rebootInProgress"
```

pkg/consts/consts.go
```diff
@@ -0,0 +1,23 @@
...
+	//MachineRoleLabel contains machine role label
+	MachineRoleLabel = "machine.openshift.io/cluster-api-machine-role"
+	// MasterMachineHealthCheck contains the MachineHealthCheck name for master nodes
+	MasterMachineHealthCheck = "masters"
+	// MasterMachineDisruptionBudget contains the MachineDisruptionBudget name for master nodes
+	MasterMachineDisruptionBudget = "masters"
+	// NamespaceOpenshiftMachineAPI contains namespace name for the machine-api componenets under the OpenShift cluster
+	NamespaceOpenshiftMachineAPI = "openshift-machine-api"
```
dhellmann: OpenShift

pkg/controllers/nodereboot/nodereboot_controller.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	if rebootInProgress {
+		return reconcile.Result{}, nil
+	}
+
+	// Creates new machine remediation object
+	mr := &mrv1.MachineRemediation{
+		ObjectMeta: metav1.ObjectMeta{
+			GenerateName: "remediation-",
```
dhellmann: Is there any reason to support having more than one remediation for a given machine at a time? If the name is set to the machine name, then you'd have free collision detection and avoid duplicates.

pkg/apis/machineremediation/v1alpha1/machineremediation_types.go
```diff
@@ -0,0 +1,74 @@
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// RemediationType contains type of the remediation
+type RemediationType string
```
zaneb: I didn't pick up on this in the spec review (perhaps because the names are almost the exact opposite of what each controller does), but after reading through all of the code I realised something.
The proposal in metal3-io/metal3-docs#48 is to implement a CRD in the baremetal-operator that will allow us to safely do a reboot without having to think about any other actors in the system. Once that exists, virtually all of the logic in RemediationTypeReboot can go away and be replaced by simply creating one of these new RebootRequest CRs.
I assumed that when that happens, this whole controller would be able to disappear, otherwise we have *3* CRs involved to do one reboot request. But that isn't your plan. This CRD will stick around as the abstraction layer for different Machine types, and the NodeReboot, which is just a controller not a CRD itself, is the thing that translates from an annotation on a Node to this Machine-level abstraction.

It does seem a little weird to me that we're saying that we need an abstraction at the level of the Machine API, and we're prototyping it in metal³, because different types of Machines (baremetal vs. cloud) need different types of remediation. But then the actual implementation lets something *else* choose the type of remediation. Shouldn't the MachineActuator (in future) or the Remediator (in this prototype) choose what type of remediation is required for the particular machine type?

pkg/controllers/nodereboot/nodereboot_controller.go
```diff
@@ -0,0 +1,141 @@
...
+		return reconcile.Result{}, err
+	}
+
+	rebootInProgress, err := isRebootInProgress(r.client, machine.Name)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	if rebootInProgress {
+		return reconcile.Result{}, nil
```
zaneb: If a reboot is in progress you actually want to check back in again later in case it has timed out, don't you?

pkg/consts/consts.go
beekhof: Not that I can see:
       machine.openshift.io/machine: openshift-machine-api/lifecycle-worker-0-f2lqd

pkg/consts/consts.go
beekhof: We can change this, but we'll need to co-ordinate with the OCP Cloud folks to make sure they set the new value

pkg/consts/consts.go
beekhof: Urgh, how do we make this distribution agnostic?

pkg/baremetal/remediator/remediator.go
```diff
@@ -0,0 +1,322 @@
...
+		if !bmh.Spec.Online && !rebootInProgress {
+			glog.V(4).Infof("Skip the remediation, machine %q has power off state before the remediation action", machine.Name)
+			bmr.recorder.Eventf(
+				machine,
+				corev1.EventTypeNormal,
+				"MachineRemediationSkippedOffline",
+				"Remediation of machine %q skipped because it was in power off state already",
+				machine.Name,
+			)
+			mrCopy.Status.State = mrv1.RemediationStateSucceeded
```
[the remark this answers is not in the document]
beekhof: Good point

pkg/apis/machineremediation/v1alpha1/doc.go
```diff
@@ -0,0 +1,4 @@
+// Package v1alpha1 contains API Schema definitions for the healthchecking v1alpha1 API group
```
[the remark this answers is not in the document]
beekhof: Yep... github.com/openshift/machine-api-operator/pkg/apis/healthchecking/

pkg/apis/machineremediation/v1alpha1/machineremediation_types.go
beekhof: I agree that some controller renaming is definitely called for.

In my ideal world MachineHealthCheck would know about power based remediation methods, but they seem adamant about not doing that.  Once https://github.com/metal3-io/metal3-docs/pull/48 is implemented, I also anticipate that the controller that implements reboot goes away.  I do expect however, that we will want some improved negative flow handling though, which might involve: try reboot N times, then escalate to reprovisioning M times, then leave it unprovisioned if even that doesn't help.  I don't believe there is a strong opinion whether that logic should belong in the controller that currently handles reboots (as the code currently implies) or elsewhere, but it is the future echos of that capability that explains why the Remediator is able to choose which operation to perform.

pkg/apis/machineremediation/v1alpha1/machineremediation_types.go
```diff
@@ -0,0 +1,74 @@
...
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []MachineRemediation `json:"items"`
+}
+
+// MachineRemediationSpec defines the spec of MachineRemediation
+type MachineRemediationSpec struct {
+	// Type contains the type of the remediation
+	Type RemediationType `json:"type,omitempty" valid:"required"`
+	// MachineName contains the name of machine that should be remediate
+	MachineName string `json:"machineName,omitempty" valid:"required"`
```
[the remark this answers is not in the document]
beekhof: Workers get new Machine names each time, but masters aren't created by MachineSets, so I agree we should key off of UUID, but perhaps keep MachineName for readability

pkg/baremetal/remediator/remediator.go
```diff
@@ -0,0 +1,322 @@
...
+	bmhCopy := bmh.DeepCopy()
+
+	// Copy the MachineRemediation object to prevent modification of the original one
+	mrCopy := machineRemediation.DeepCopy()
+
+	now := time.Now()
+	switch machineRemediation.Status.State {
+	// initiating the reboot action
+	case mrv1.RemediationStateStarted:
+		rebootInProgress := isRebootInProgress(bmh)
```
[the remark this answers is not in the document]
beekhof: Doug's suggestion is to keep everything required for the state machine in the Host object (stuffed into an annotation or two).  That should kill a lot of racing possibilities.

pkg/apis/machineremediation/v1alpha1/machineremediation_types.go
beekhof: I believe .Reason relates to the current value of .State rather than why the remediation was started.
We wouldn't normally expect the user to get involved, and there is no mechanism for machine healthcheck to pass that via the annotation.

pkg/baremetal/remediator/remediator.go
beekhof: Correct.  Or even if the power is successfully restored, the machine may remain borked.
This is looking ahead to when we want to escalate to reprovisioning when reboot fails.

pkg/baremetal/remediator/remediator.go
beekhof: Possibly this could be something that is written to .Status.Reason

pkg/apis/machineremediation/v1alpha1/machineremediation_types.go
beekhof: Agreed. As discussed we would like to address this after thursday :)

pkg/baremetal/remediator/remediator.go
beekhof: Correct, garbage collection.
This relates to your other question about multiple remediation objects and is something that will likely affect metal3-io/metal3-docs#48 too.  I'll followup in the "multiple remediation objects" thread

pkg/components/deployments.go
beekhof: yep, we can kill this file.

pkg/components/rbac.go
beekhof: We need permissions to where the machines live.
Is there a better way to express this in a distro agnostic manner?  Maybe move it to a manifest?

beekhof:
> > I'm not sure if we want to land this upstream in the v1alpha1 branch at all.
> 
> I think it makes most sense to merge it where it can best be tested first, which would be in v1alpha1. Once it's solid, porting it forward to master should hopefully not be that difficult.

I believe Doug was suggesting we kill this PR (and possibly the v1alpha1 branch) and instead work in https://github.com/openshift/cluster-api-provider-baremetal/pull/47 before porting to master in metal3-io

pkg/controllers/nodereboot/nodereboot_controller.go
```diff
@@ -0,0 +1,141 @@
...
+	if err = r.client.Create(context.TODO(), mr); err != nil {
+		return reconcile.Result{}, err
+	}
+	return reconcile.Result{}, nil
+}
+
+func isRebootInProgress(c client.Client, machineName string) (bool, error) {
+	machineRemediations := &mrv1.MachineRemediationList{}
+	listOptions := &client.ListOptions{}
+	if err := c.List(context.TODO(), listOptions, machineRemediations); err != nil {
```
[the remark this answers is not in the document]
beekhof: Yep, this is tracked as https://issues.redhat.com/browse/MGMT-324 and is being actively worked on

pkg/controllers/nodereboot/nodereboot_controller.go
```diff
@@ -0,0 +1,141 @@
+package nodereboot
```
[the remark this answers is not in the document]
beekhof: It doesn't sit well in my head either.  +1 for some renaming

pkg/baremetal/remediator/remediator.go
```diff
@@ -0,0 +1,322 @@
...
+	node, err := getNodeByMachine(c, machine)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			glog.Warningf("The machine %q node does not exist", machine.Name)
+			return nil
+		}
+		return err
+	}
+
+	return c.Delete(context.TODO(), node)
```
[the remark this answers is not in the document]
beekhof: I bet it does.  We'd better check that before returning

pkg/controllers/nodereboot/nodereboot_controller.go
beekhof: yes

pkg/controllers/nodereboot/nodereboot_controller.go
beekhof: In parallel: no, but we do want some history somewhere.

If we went back to storing the last N remediation events in the CR's .Status, then we could have one per Machine, with the same name as the Machine as you suggest.

We'd need to be careful about new remediation requests happening around about the time the previous one was wrapping up, but it seems doable.

pkg/baremetal/remediator/remediator.go
beekhof: In theory an admin or any subsystem that deletes Machines or knows about BMH's could have triggered it.

pkg/apis/machineremediation/v1alpha1/machineremediation_types.go
zaneb: Props for the Red Dwarf reference :)

pkg/baremetal/remediator/remediator.go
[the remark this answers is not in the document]
zaneb: Yes, having the read and write on the same object is another way to ensure you can't change the state based on a stale read. (I'm not a big fan of annotations because there's no RBAC for them like there is for the Status of a controller's own CRD.)

zaneb:
/hold
We're currently rethinking this plan to go with something simpler + implement the reboot functionality in the BaremetalHost straight away rather than start with a temporary implementation of that part here.

dhellmann:
We are going to close the v1alpha1 branch in this repository, so I am closing open PRs that are no longer needed.

This PR has moved to https://github.com/openshift/cluster-api-provider-baremetal/pull/47 and a version will be brought up to the master branch in this repository later.

PR #226, opened 2020-01-22
Define release guidelines and update Makefile.

Arvinderpal:
We still need to address the following:

- [x] Have the CI build the image automatically and push them to quay.io
- [x] Associate images on quay.io with releases on github? 
- [x] Creating a release on github using their web UI
- [ ] [nice-to-have) Create releases automatically based on git tags. 
- [ ] Investigate multi-arch build support with quay.io

Issue:
https://github.com/metal3-io/cluster-api-provider-baremetal/issues/224

@maelk @russellb

### DIFF
--- a/docs/releasing.md
+++ b/docs/releasing.md
@@ -1,0 +1,53 @@
+# Releasing
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+
+
+- [Prerequisites](#prerequisites)
+  - [`gcloud`](#gcloud)
+  - [`docker`](#docker)
+- [Output](#output)
+  - [Expected artifacts](#expected-artifacts)
+  - [Artifact locations](#artifact-locations)
+- [Process](#process)
+  - [Permissions](#permissions)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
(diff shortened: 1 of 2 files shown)


docs/releasing.md
```diff
@@ -0,0 +1,53 @@
...
+### Expected artifacts
+
+1. A container image of the shared cluster-api-provider-baremetal manager
+2. A git tag 
+3. Manifest file - `infrastructure-components.yaml`
+
+### Artifact locations
+
+1. The container image is found in the registry `quay.io/metal3-io` with an image
+   name of `cluster-api-provider-baremetal-<ARCH>` and a tag that matches the release version.
```
maelk: right now we do not have a multi-arch setup. We need to investigate this with quay.io

maelk:
On quay.io the image is built directly with the correct tag, and updated when needed, so I would say that the second item is completed too. For the third one, I just did the first release with the web UI. It is also very straightforward to do it  with the git command line (though the asset upload is problematic) or with some scripts. There are many existing already.

Arvinderpal:
I see the `0.2.0` release - Great!
I'm okay with using the web UI. If the need arises, we can consider alternatives.

docs/releasing.md
Arvinderpal: `arm` support is a must.

maelk:
This was merged as part of #266

PR #250, opened 2020-02-14
✨ Add default, validation and conversion webhooks

maelk:
This PR adds for BareMetalCluster, BareMetalMachine and BareMetalMachineTemplate:
- the version conversion functions
- conversion webhooks
- default webhooks
- validation webhooks 

It also upgrade the CRD version to v1, and controller-gen to v0.2.5


baremetal/baremetalmachine_manager.go
```diff
@@ -36,10 +36,10 @@ import (
 	"k8s.io/utils/pointer"
 
 	bmh "github.com/metal3-io/baremetal-operator/pkg/apis/metal3/v1alpha1"
+	capbm "github.com/metal3-io/cluster-api-provider-baremetal/api/v1alpha3"
```
kashifest: I think we should use infrav1 in all places to be more consistent.

api/v1alpha3/baremetalmachineTemplate_webhook_test.go
```diff
@@ -0,0 +1,91 @@
...
+	c := &BareMetalMachineTemplate{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: "fooboo",
+		},
+		Spec: BareMetalMachineTemplateSpec{},
+	}
+	c.Default()
+}
+
+func TestBareMetalMachineTemplateValidation(t *testing.T) {
```
kashifest: Should we only check Image for validation? No other fields to check?

PR #259, opened 2020-02-20
✨ Fulfill CAPI provider contract

maelk:
This PR adds the needed features to fulfill CAPI contract


maelk:
/hold
needs proper description and further work

PR #263, opened 2020-02-21
✨ Set ownerreference on BareMetalHost

maelk:
This PR adds an Ownerreference on the baremetalhost, set by CAPBM when selecting the baremetalhost. Clusterctl builds a tree of resources using the ownerreferences. the ownerreference is not directly used right now because we still have BMO installed out of the scope of clusterctl, but whenever that will not be the case anymore (i.e. we include BMO deployment in infrastructure-components.yaml) then this ownerreference is a must to ensure that the BMH gets moved along during the pivoting.

### DIFF
--- a/baremetal/baremetalmachine_manager.go
+++ b/baremetal/baremetalmachine_manager.go
@@ -346,6 +346,7 @@ func (m *MachineManager) Delete(ctx context.Context) error {
 	}
 
 	if host != nil && host.Spec.ConsumerRef != nil {
+		host.OwnerReferences = m.DeleteOwnerRef(host.OwnerReferences)
 		// don't remove the ConsumerRef if it references some other bare metal machine
 		if !consumerRefMatches(host.Spec.ConsumerRef, m.BareMetalMachine) {
 			m.Log.Info("host already associated with another bare metal machine",
@@ -455,6 +456,15 @@ func (m *MachineManager) Update(ctx context.Context) error {
 	}
 	if host == nil {
 		return fmt.Errorf("host not found for machine %s", m.Machine.Name)
+	}
+
+	// ensure that the BMH specs are correctly set
+	err = m.setHostSpec(ctx, host)
+	if err != nil {
+		m.setError("Failed to associate the BaremetalHost to the BareMetalMachine",
+			capierrors.CreateMachineError,
+		)
+		return err
 	}
 
 	err = m.ensureAnnotation(ctx, host)
@@ -685,6 +695,8 @@ func (m *MachineManager) setHostSpec(ctx context.Context, host *bmh.BareMetalHos
 	}
 
 	host.Spec.Online = true
+	// Set OwnerReferences
+	host.OwnerReferences = m.SetOwnerRef(host.OwnerReferences, true)
 	return m.client.Update(ctx, host)
 }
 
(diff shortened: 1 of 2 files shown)


dhellmann:
Could we get a little more detail about why the existing consumerRef field is insufficient for this?

I've noticed some other patches merging rather quickly this morning. I would like to have time to review this one, so I'm applying a temporary hold.

/hold

maelk:
Sure, this is because clusterctl builds a tree of resources using the ownerreferences. It is not directly used right now because we still have BMO installed out of the scope of clusterctl, but whenever that will not be the case anymore (i.e. we include BMO deployment in infrastructure-components.yaml) then this ownerreference is a must to ensure that the BMH gets moved along during the pivoting. It is duplicating the idea behind consumerRef, but clusterctl does not understand BMH CRs, so it uses a more universal way to build the dependency tree.

baremetal/baremetalmachine_manager.go
```diff
@@ -346,6 +346,7 @@ func (m *MachineManager) Delete(ctx context.Context) error {
 	}
 
 	if host != nil && host.Spec.ConsumerRef != nil {
+		host.OwnerReferences = m.DeleteOwnerRef(host.OwnerReferences)
```
dhellmann: For the ConsumerRef, we check that it matches the current machine before making any changes. Should we do the same here?

baremetal/baremetalmachine_manager.go
```diff
@@ -837,3 +849,53 @@ func (m *MachineManager) SetProviderID(providerID string) {
...
+	return refList
+}
+
+// DeleteOwnerRef removes the ownerreference to this baremetal machine
+func (m *MachineManager) DeleteOwnerRef(refList []metav1.OwnerReference) []metav1.OwnerReference {
+	if len(refList) == 0 {
+		return refList
+	}
+	index, err := m.FindOwnerRef(refList)
+	if err == nil {
```
dhellmann: This logic seems inverted from what I'm used to seeing in go. If this line was `if err != nil` then the block could return immediately, which would place the shortcut return path closer to the test and keep the more common active path in the main block of the function.

baremetal/baremetalmachine_manager.go
```diff
@@ -837,3 +849,53 @@ func (m *MachineManager) SetProviderID(providerID string) {
...
+		refList[index] = refList[refListLen]
+		return (m.DeleteOwnerRef(refList[:refListLen-1]))
+	}
+	return refList
+}
+
+// FindOwnerRef checks if an ownerreference to this baremetal machine exists
+// and returns the index
+func (m *MachineManager) FindOwnerRef(refList []metav1.OwnerReference) (int, error) {
+	for i := 0; i < len(refList); i++ {
```
dhellmann: Is there a reason not to use range() here?

```suggestion
	for i, curOwnerRef := range(refList) {
```

baremetal/baremetalmachine_manager_test.go
```diff
@@ -1809,18 +1812,25 @@ var _ = Describe("BareMetalMachine manager", func() {
 			} else {
 				_, ok := errors.Cause(err).(HasRequeueAfterError)
 				Expect(ok).To(BeTrue())
+				return
```
dhellmann: This looks like it's going to short circuit some of the test logic. Is that the intent? Why return early when a re-queue is expected? Perhaps this is another place where the logic could be flipped and the return kept inside the if/then block and the else block eliminated.

baremetal/baremetalmachine_manager.go
maelk: The DeleteOwnerRef will only remove the ownerRef pointing to this baremetalmachine only. It will leave the others intact

baremetal/baremetalmachine_manager.go
maelk: good point! I'll fix it

baremetal/baremetalmachine_manager.go
maelk: no, indeed. I will fix it

mhrivnak:
A main reason we did not add such an owner reference to BMH previously was that it does not make sense to cascade-delete a BMH. Is that accounted for somehow with this change?

baremetal/baremetalmachine_manager_test.go
maelk: indeed, I have tightened the condition for the return

maelk:
When deleting the Baremetalmachine, capbm will block the deletion with a finalizer until the BMH is back to "ready" state. The first operation of the deprovisioning is to remove the  ownerReference. So when the Baremetalmachine is garbage collected, the BMH does not have an ownerreference to the baremetalmachine anymore and is not deleted. Is this sufficient in your opinion, or do you foresee corner-cases ?

dhellmann:
> When deleting the Baremetalmachine, capbm will block the deletion with a finalizer until the BMH is back to "ready" state. The first operation of the deprovisioning is to remove the ownerReference. So when the Baremetalmachine is garbage collected, the BMH does not have an ownerreference to the baremetalmachine anymore and is not deleted. Is this sufficient in your opinion, or do you foresee corner-cases ?

So when a BMM is deleted, first the owner reference is removed and then deprovisioning starts? And after deprovisioning is done then the finalizer is removed from the BMM and it is deleted?

maelk:
Yes, that's the workflow

mhrivnak:
> When deleting the Baremetalmachine, capbm will block the deletion with a finalizer until the BMH is back to "ready" state. The first operation of the deprovisioning is to remove the ownerReference. So when the Baremetalmachine is garbage collected, the BMH does not have an ownerreference to the baremetalmachine anymore and is not deleted. Is this sufficient in your opinion, or do you foresee corner-cases ?

Sounds reasonable to me. Thanks for the explanation.

kashifest:
Looks good to me

PR #264, opened 2020-02-21
✨ Modify the userdata secret creation logic

maelk:
Currently, we always create a secret with the userdata for BMH. Since CABPK v1alpha3, CABPK creates a secret containing the userData. This PR removes the duplication of the secret where possible.

Several cases:

- no DataSecretName set, but Data is set in the Machine Bootstrap -> creates the secret
- DataSecretName set, but Machine and BMH are in different namespaces -> creates the secret, since the secret is referenced by name only in both cases, assuming that it is in the same namespace as the machine or as the BMH
- DataSecretName set and Machine and BMH in same namespace -> re-use CABPK secret

### DIFF
--- a/baremetal/baremetalmachine_manager.go
+++ b/baremetal/baremetalmachine_manager.go
@@ -262,30 +262,22 @@ func (m *MachineManager) Associate(ctx context.Context) error {
 
 // Merge the UserData from the machine and the user
 func (m *MachineManager) GetUserData(ctx context.Context) error {
-	var err error
-	var decodedUserDataBytes []byte
 	// if datasecretname is set get userdata from secret
 	if m.Machine.Spec.Bootstrap.DataSecretName != nil {
-		capiBootstrapSecret := corev1.Secret{}
-		capikey := client.ObjectKey{
+		m.BareMetalMachine.Spec.UserData = &corev1.SecretReference{
 			Name:      *m.Machine.Spec.Bootstrap.DataSecretName,
 			Namespace: m.Machine.Namespace,
 		}
-		err := m.client.Get(ctx, capikey, &capiBootstrapSecret)
-		if err != nil {
-			return err
-		}
-		decodedUserDataBytes = capiBootstrapSecret.Data["value"]
-	} else if m.Machine.Spec.Bootstrap.Data != nil {
-		// if datasecretname is not set
-		decodedUserData := *m.Machine.Spec.Bootstrap.Data
-		// decode the base64 cloud-config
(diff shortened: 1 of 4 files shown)


baremetal/baremetalmachine_manager.go
```diff
@@ -426,36 +444,43 @@ func (m *MachineManager) Delete(ctx context.Context) error {
...
-			m.setError("Failed to delete BareMetalMachine",
-				capierrors.DeleteMachineError,
-			)
-			return err
+	// Delete created secret, if data was set without DataSecretName or if
+	// BareMetalHost and Machine are in different namespaces.
+	if (m.Machine.Spec.Bootstrap.DataSecretName == nil &&
+		m.Machine.Spec.Bootstrap.Data != nil) ||
+		(m.Machine.Spec.Bootstrap.DataSecretName != nil &&
+			m.Machine.Namespace != host.Namespace) {
```
kashifest: This seems a very complex logic, can we simplify it?

baremetal/baremetalmachine_manager.go
```diff
@@ -426,36 +444,43 @@ func (m *MachineManager) Delete(ctx context.Context) error {
...
 				capierrors.DeleteMachineError,
 			)
 			return err
+		} else if err == nil {
+			//unset the finalizers (remove all since we do not expect anything else
+			// to control that object)
+			tmpBootstrapSecret.Finalizers = []string{}
+			err = m.client.Update(ctx, &tmpBootstrapSecret)
+			if err != nil {
+				m.setError("Failed to delete BareMetalMachine",
```
kashifest: Should the message contain some info about the actual reason i.e. cannot delete Bootstrap secret, Failed to delete BMM or something like that?

baremetal/baremetalmachine_manager.go
```diff
@@ -426,36 +444,43 @@ func (m *MachineManager) Delete(ctx context.Context) error {
...
+					capierrors.DeleteMachineError,
+				)
+				return err
+			}
+			// Delete the secret with use data
+			err = m.client.Delete(ctx, &tmpBootstrapSecret)
+			if err != nil {
+				m.setError("Failed to delete BareMetalMachine",
```
kashifest: Should the message contain some info about the actual reason i.e. cannot delete Bootstrap secret, Failed to delete BMM or something like that?

baremetal/baremetalmachine_manager.go
maelk: not really. There are two cases, the first one is that objects are not in the same namespace when DataSecretName is set, and the second one is that DataSecretName is not set but Data is set. I don't see how to make it simpler.

